PR #34, opened 2024-07-03
feat: Added Support for Corona Renderer

maxschramp:

Only Default and ART renderers were supported, rendering a file with Corona would default to the scanline renderer.

I duplicated the ART render handler and adjusted it to be a Corona Handler. I also adjusted the allowed output files to include .cxr (Corona EXR Format) and the allowed renderers to include Corona

Minor impact, should just add support for Corona.

I built the wheel and installed it and was able to run Corona jobs.

Only through commit messages, based on the format of the surrounding code additional comments were unnecessary.

No

----

### DIFF
--- a/src/deadline/max_adaptor/MaxAdaptor/schemas/init_data.schema.json
+++ b/src/deadline/max_adaptor/MaxAdaptor/schemas/init_data.schema.json
@@ -19,14 +19,16 @@
                 ".rpf",
                 ".tga",
                 ".tif",
-                ".dds"
+                ".dds",
+                ".cxr"
             ]
         },
         "state_set": { "type": "string" },
         "renderer": {
             "enum": [
                 "Default_Scanline_Renderer",
-                "ART_Renderer"
+                "ART_Renderer",
+                "Corona"
             ]
         },
         "scene_file": { "type": "string" },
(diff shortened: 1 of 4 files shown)


maxschramp:
Hey, this is my first PR ever. Realized I did the Semantic and DCO stuff wrong but let me know if there's anything else I can do to help as I see it needs testing on the different versions of Python?

epmog:
Thanks for the pull request, it's super appreciated! Congrats on it being your first! 🎊 

Yea the DCO/Semantic seems to be a painpoint for the first few contributions so maybe we can do something better to help people get it right from the beginning (or when coming back).

I think the linting will fail due to some formatting, but I'll approve and run it so you can see. The actions mostly do something like this that you can replicate locally:

```
hatch run fmt
hatch run lint
hatch build
hatch run test
```

epmog (dismissed):

lucaseck (dismissed):

maxschramp:
Thanks for the tips! I think the DCO/Semantic stuff makes sense, I just didn't realize it until the the checks showed up here. I ran the actions, I think the only problem was I missed a comma. Hopefully this works!

epmog (approved):

lucaseck (approved):

epmog:
Thanks again for the pull request, this is now merged and will be available in the next release!

maxschramp:
Awesome thanks @epmog! Looking forward to contributing more in the future!

Issue #33, opened 2024-06-28
Feature request: Vray RENDERERS


mmasdoria:
### Use Case

I want to use vray renderer in 3dsma but it seems not allowed yet in deadline-cloud-for-3ds-max.
I can see in data_const.py : 
 ALLOWED_RENDERERS = ["Default_Scanline_Renderer", "ART_Renderer"]

### Proposed Solution

Do you know when Vray renderer will be available ?

ddneilson:
Thank you for the interest in our product, and taking the time to open a feature request. We'll add this to our list of issues to prioritize, but we cannot provide an estimate for when this work might be taken up by our team. In the meantime, we'd be happy to accept a pull request from the community that adds this functionality.

maxschramp:
I was able to add Corona do my data_const.py file but I've yet to successfully render with it as I am still setting up my AWS farm. I may have some free time in the next week or two to delve deeper into it and put together a PR.

maxschramp:
I've been thinking about this for a couple weeks, when I get the current renderer using `rt.renderers.current` in 3dsMax Python it shows up as `V_Ray_6__update_2_1`. I'm assuming that other V-Ray versions will show up in a similar format with the version formatting at the end, I'm going to see if there's a good way to do this without adding multiple V-Ray versions to the dictionary but wanted to update the thread with this context.

maxschramp:
@mmasdoria merged!

jairaws:
Support for V-Ray renders released in v0.1.2.
https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/64

PR #35, opened 2024-07-17
feat: Added Support for V-Ray 6 and V-Ray GPU 6 Renderers

maxschramp:
V-Ray and V-Ray GPU renderers didn't have render handlers and failed sanity checks. #33 
Created a render handler for V-Ray that can be switched to V-Ray GPU with a boolean when initializing it. Currently only supports V-Ray 6 major version as any earlier versions don't support 3dsMax 2024, can be re-evaluated when V-Ray 7 is released.
Several places I added a `.Split("__")` to the renderer class names to split off the minor version. `V_Ray_6__update_2_1 `is split into `V_Ray_6`. This shouldn't have any effect on other render engines as they don't have a double underscore in the name, but allows all V-Ray 6 minor versions to be supported without having to define every version in the `ALLOWED_RENDERERS` constant. If it's preferred to do it that way though - I'm happy to update.
Was able to successfully submit, initialize, and run a job locally using 3dsMax 2024. Submitter and Adaptor worked as expected with the new handlers. Ran hatch formatting, linting, and tests locally.
Inline Comments
No
----

### DIFF
--- a/src/deadline/max_adaptor/MaxAdaptor/schemas/init_data.schema.json
+++ b/src/deadline/max_adaptor/MaxAdaptor/schemas/init_data.schema.json
@@ -2,9 +2,15 @@
     "$schema": "https://json-schema.org/draft/2020-12/schema",
     "type": "object",
     "properties": {
-        "camera": { "type": "string" },
-        "output_file_path": { "type": "string" },
-        "output_file_name": { "type": "string" },
+        "camera": {
+            "type": "string"
+        },
+        "output_file_path": {
+            "type": "string"
+        },
+        "output_file_name": {
+            "type": "string"
+        },
         "output_file_format": {
             "enum": [
                 ".avi",
@@ -24,18 +30,30 @@
                 ".vrimg"
             ]
         },
-        "state_set": { "type": "string" },
+        "state_set": {
+            "type": "string"
+        },
         "renderer": {
-            "enum": [
-                "Default_Scanline_Renderer",
-                "ART_Renderer",
-                "Corona",
-                "V_Ray_6",
-                "V_Ray_GPU_6"
+            "oneOf": [
+                {
+                    "type": "string",
+                    "pattern": "^V_Ray_[1-9](_GPU)?.*$"
+                },
+                {
+                    "enum": [
+                        "Default_Scanline_Renderer",
+                        "ART_Renderer",
+                        "Corona"
+                    ]
+                }
             ]
         },
-        "scene_file": { "type": "string" },
-        "strict_error_checking": { "type": "boolean" }
+        "scene_file": {
+            "type": "string"
+        },
+        "strict_error_checking": {
+            "type": "boolean"
+        }
     },
     "required": [
         "renderer",
--- a/src/deadline/max_adaptor/MaxAdaptor/schemas/init_data.schema.json
+++ b/src/deadline/max_adaptor/MaxAdaptor/schemas/init_data.schema.json
@@ -20,7 +20,8 @@
                 ".tga",
                 ".tif",
                 ".dds",
-                ".cxr"
+                ".cxr",
+                ".vrimg"
             ]
         },
         "state_set": { "type": "string" },
@@ -28,7 +29,9 @@
             "enum": [
                 "Default_Scanline_Renderer",
                 "ART_Renderer",
-                "Corona"
+                "Corona",
+                "V_Ray_6",
+                "V_Ray_GPU_6"
             ]
         },
         "scene_file": { "type": "string" },
--- a/src/deadline/max_adaptor/MaxClient/render_handlers/__init__.py
+++ b/src/deadline/max_adaptor/MaxClient/render_handlers/__init__.py
@@ -3,6 +3,7 @@
 from .default_max_handler import DefaultMaxHandler
 from .art_handler import ArtHandler
 from .corona_handler import CoronaHandler
+from .vray_handler import VrayHandler
 
 __all__ = ["DefaultMaxHandler", "get_render_handler"]
 
@@ -21,5 +22,9 @@ def get_render_handler(renderer: str = "Default_Scanline_Renderer") -> DefaultMa
         return ArtHandler()
     elif renderer == "Corona":
         return CoronaHandler()
+    elif renderer == "V_Ray_6":
+        return VrayHandler(GPU=False)
+    elif renderer == "V_Ray_GPU_6":
+        return VrayHandler(GPU=True)
     else:
         return DefaultMaxHandler()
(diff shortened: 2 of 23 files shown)


lucaseck:
Thank you for the second pull request @maxschramp! Great to see after the previous additions for Corona. We will start taking a look and review the changes.

maxschramp:
Thanks @lucaseck, Looking forward to any feedback!

amarsjac (dismissed):
Looks good to me!

src/deadline/max_adaptor/MaxClient/render_handlers/vray_handler.py
```diff
@@ -0,0 +1,51 @@
...
+    def check_renderer(self) -> None:
+        """
+        Checks if the active renderer is set to V-Ray. If it is not, set it to the latest version V-Ray.
+        Gets the latest versions of V-Ray and V-Ray GPU from rt.rendererclass.classes.
+        """
+        current_renderer = str(rt.renderers.current).split(":")[0].split("__")[0]
+
+        # The V-Ray renderer class name is "V_Ray_6__update_#_#" and "V_Ray_GPU_6__update_#_#"
+        vray = [
+            i for i in list(rt.rendererclass.classes) if "V_Ray" in str(i) and "GPU" not in str(i)
```
lucaseck: With 3DSMax2024 is there always a V_Ray* render class available that would match both of these searches or does a VRay plugin need to be installed for them to exist? Might be good to do a check and ensure that a vray and vray_gpu render class is found at all before taking [-1] and raise an exception if these searches don't find a matching render class.

src/deadline/max_submitter/data_const.py
```diff
@@ -37,6 +43,7 @@
     ["TIF Image File (*.tif)", ".tif"],
     ["DDS Image File (*.dds)", ".dds"],
     ["Corona EXR Image Format (*.cxr)", ".cxr"],
+    ["V-Ray image formamt (*.vrimg)", ".vrimg"],
```
lucaseck: Spelling nit: `V-Ray image formamt (*.vrimg)` -> `V-Ray Image Format (*.vrimg)`

src/deadline/max_adaptor/MaxAdaptor/schemas/init_data.schema.json
```diff
@@ -20,15 +20,18 @@
...
                 ".tif",
                 ".dds",
-                ".cxr"
+                ".cxr",
+                ".vrimg"
             ]
         },
         "state_set": { "type": "string" },
         "renderer": {
             "enum": [
```
lucaseck: Another option that may help better support more versions of VRay I believe is to do something like:
```
"renderer": {
    "oneOf": [
        {
            "type": "string",
            "pattern": "^V_Ray_[1-9](_GPU)?.*$"
        }
        {
            "enum": [
                "Default_Scanline_Renderer",
                "ART_Renderer",
                "Corona",
            ]
        }
    ]
}
```
No need to update for now unless you're interested in exploring that option, but it's something that might come up in the future. That would also allow setting the renderer to the full versioned renderer and not have to split at the "__". Then the `.split("__")` would only be needed when checking against the ALLOWED_RENDERERS list and otherwise it could be stored and displayed as the full string I believe.

src/deadline/max_adaptor/MaxClient/render_handlers/vray_handler.py
maxschramp: Good call!

src/deadline/max_adaptor/MaxAdaptor/schemas/init_data.schema.json
maxschramp: I tested this out and it worked great. Nice way to avoid all those extra splits as well.

maxschramp:
Sorry about those extra commits, forgot to run the format/linting. I adjusted per your suggestions @lucaseck, and ran a render that worked as expected. Thanks for your help, let me know if there's anything else!

src/deadline/max_adaptor/MaxClient/render_handlers/__init__.py
```diff
@@ -21,5 +22,9 @@ def get_render_handler(renderer: str = "Default_Scanline_Renderer") -> DefaultMa
         return ArtHandler()
     elif renderer == "Corona":
         return CoronaHandler()
+    elif renderer == "V_Ray_6":
```
lucaseck: With the removal of some of the `.strip("__")` other than checking against the allowed renderers, does the renderer not come through now as the full version string e.g. `V_Ray_6__update_#_#`? Just making sure that we don't need to be doing a `renderer.startswith()` for the VRay cases.

src/deadline/max_adaptor/MaxClient/render_handlers/vray_handler.py
```diff
@@ -0,0 +1,64 @@
+"""
+3ds Max Deadline Cloud Adaptor - V-Ray specific actions
+
+Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+"""
+
+import sys
+
+import pymxs  # noqa
```
lucaseck: nit: as the code scanner found this unused import could be removed

src/deadline/max_adaptor/MaxClient/render_handlers/vray_handler.py
```diff
@@ -0,0 +1,64 @@
...
+
+class VrayHandler(DefaultMaxHandler):
+    """Render Handler for V-Ray"""
+
+    def __init__(self, GPU):
+        """
+        Initializes the V-Ray and V-Ray Handler
+        """
+        super().__init__()
+        self.GPU: bool = GPU
```
lucaseck: One of the code scanners is flagging this variable as not following conventions. Can the variable and input parameter be changed to all lowercase `gpu`?

lucaseck (commented):
Thanks for the updates @maxschramp just a couple of small nits and a follow-up question but otherwise it looks good to me!

src/deadline/max_adaptor/MaxClient/render_handlers/__init__.py
maxschramp: I think you're right that we'll need the` .startswith() `- let me test it out locally and I'll update.

maxschramp: I updated, it worked locally. I think since only V-Ray 6 is available for 3dsMax 2024 right now (and this project has that as a min. requirement) it is okay for now, but I think in the future it would be useful to add a `major_version` variable to the V-Ray handler so this if statement doesn't get too long.

lucaseck (approved):
LGTM 🚢, thanks @maxschramp for another contribution.

src/deadline/max_adaptor/MaxClient/render_handlers/__init__.py
lucaseck: Agreed, it can be a future improvement. No sense holding this PR more on that. Approved.

amarsjac (approved):

maxschramp:
Sweet! Thanks for helping cook this up 🔥

PR #61, opened 2024-12-11
chore(deps): Update the version of multiple dependencies.

jairaws:
We have multiple dependencies that need to be updated.

Update the version of all the relevant dependencies.

Dependencies up to date.

Submitted job and verified that succeeded.

N/A.

No
----

Screenshot 2024-12-11 at 2 26 06 PM

### DIFF
--- a/requirements-testing.txt
+++ b/requirements-testing.txt
@@ -5,6 +5,6 @@ mypy == 1.*
 pytest == 8.*
 pytest-cov == 5.*
 pytest-xdist == 3.*
-ruff == 0.6.*
+ruff == 0.7.*
 twine == 5.*
 types-pyyaml == 6.*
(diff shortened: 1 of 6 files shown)


requirements-testing.txt
```diff
@@ -1,10 +1,11 @@
...
+pytest-cov == 6.*
 pytest-xdist == 3.*
-ruff == 0.7.*
-twine == 5.*
+twine == 6.*
+mypy == 1.*
+black == 24.*
+ruff == 0.8.*
+moto[s3,sts] == 5.*
+pipgrip == 0.10.*
```
jericht: This looks like a new dependency, do we need it?

jairaws: Good catch. I copy the dependencies list from `deadline-cloud-for-maya` to have them in the same order. Didn't notice this extra one. Will remove it.

jericht (approved):

AlexTranAmz (approved):

PR #72, opened 2025-01-23
feat: Add support for 3dsmaxbatch

jairaws:
We want the 3dsMax adaptor to be able to use 3dsmaxbatch. However, when trying to use it, the adaptor would freeze without marking the tasks as complete.

* Create MaxExecutableHandler to help support multiple 3dsMax executables.
* Create MaxRegexCallback to take care of removing unexpected characters from 3dsaxbatch.
* Update the progress regex to match the output from 3dsmaxbatch.
* Implement log_to_console() method in the DefaultMaxHandler class, to use a specific logging strategy in function of what 3dsMax executable is being used.

3dsmaxbatch is supported.

* Unit tests.
* Submitted jobs and validate they succeeded.

No
----

Screenshot 2025-01-23 at 10 26 32 AM

### DIFF
--- a/src/deadline/max_adaptor/MaxAdaptor/regex_callback_handler.py
+++ b/src/deadline/max_adaptor/MaxAdaptor/regex_callback_handler.py
@@ -1,0 +1,25 @@
+"""
+3ds Max Deadline Cloud Adaptor - 3dsMax Regex Callback Handler
+
+Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+"""
+
+from __future__ import annotations
+import re
+from openjd.adaptor_runtime.app_handlers import RegexCallback
+
+
+class MaxRegexCallback(RegexCallback):
+
+    def __init__(
+        self, regex_list, callback, exit_if_matched=False, only_run_if_first_matched=False
+    ):
+        super().__init__(regex_list, callback, exit_if_matched, only_run_if_first_matched)
+
+    def get_match(self, msg: str) -> re.Match | None:
+        """
+        Takes care of a special case in 3dsMax, in which unexpected characters are added to executable output.
+        The rest of the functionality is the as in the parent class.
+        :param msg: The text to parse.
+        """
+        return super().get_match(msg.replace("\x00", ""))
--- a/src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
+++ b/src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
@@ -13,6 +13,8 @@ import sys
 import pymxs  # noqa
 from pymxs import runtime as rt
 
+from deadline.max_adaptor.executable_handler import MaxExecutableHandler, SupportedMaxExecutable
+
 logger = logging.getLogger(__name__)
 
 # Re-assign sys stdout and stderr to print in the console instead of the Max Listener
@@ -37,6 +39,7 @@ class DefaultMaxHandler:
         self.output_dir = None
         self.output_name = None
         self.output_format = None
+        self._executable_handler: MaxExecutableHandler = MaxExecutableHandler()
 
     def start_render(self, data: dict) -> None:
         """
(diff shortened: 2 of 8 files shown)


src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
```diff
@@ -264,11 +269,21 @@ def set_scene_file(self, data: dict):
...
+
+    def log_to_console(self, message: str) -> None:
+        """
+        Handles logging to the stdout, based on which 3dsMax executable is being used.
+        :param message: The text to log to the stdout.
+        """
+        if self._executable_handler.is_executable_type(SupportedMaxExecutable.BATCH):
+            rt.logsystem.logEntry(message, broadcast=True)
+        else:
+            print(message, flush=True)
```
sakshie95, line 289: Why is ``flush`` True for non batch executable type? Do we not want to wait for the lines to be complete before printing them intentionally?

jairaws: No. We want to get the logs out asap, because we are consuming them in real time, as events. The sooner we get them, the sooner we finish a task.

src/deadline/max_adaptor/MaxAdaptor/regex_callback_handler.py
```diff
@@ -0,0 +1,25 @@
...
+    ):
+        super().__init__(regex_list, callback, exit_if_matched, only_run_if_first_matched)
+
+    def get_match(self, msg: str) -> re.Match | None:
+        """
+        Takes care of a special case in 3dsMax, in which unexpected characters are added to executable output.
+        The rest of the functionality is the as in the parent class.
+        :param msg: The text to parse.
+        """
+        return super().get_match(msg.replace("\x00", ""))
```
leongdl, line 25: I think \x00 is a null string terminator, I'm surprised this borks the regex parsing. 

Is it only the last character of a string? Or are there any occurrences within the string?

src/deadline/max_adaptor/executable_handler.py
```diff
@@ -0,0 +1,81 @@
...
+from enum import Enum
+from os import environ
+from functools import cache
+import re
+from typing import List, Dict, Match, Optional
+from dataclasses import dataclass
+
+
+class SupportedMaxExecutable(Enum):
+    INTERACTIVE = "3dsmax"
```
leongdl, line 16: Interesting change, this will be great for the other project I'm working on.

src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
```diff
@@ -264,11 +269,21 @@ def set_scene_file(self, data: dict):
...
             raise FileNotFoundError(f"Error: The scene file '{file_path}' does not exist")
         try:
             rt.SetQuietMode(True)
             rt.loadMaxFile(file_path, quiet=True)
         except Exception:
-            print(f"Error: while opening '{file_path}'")
+            self.log_to_console(f"Error: while opening '{file_path}'")
             raise RuntimeError(f"Error: while opening '{file_path}'")
+
+    def log_to_console(self, message: str) -> None:
```
leongdl, line 281: Would this method be better as a member of `MaxExecutableHandler`? It can encapsulate all the if-else logic between the different exes. Then we can do `self._executable_handler.log....`

src/deadline/max_adaptor/MaxAdaptor/regex_callback_handler.py
jairaws: Many occurrences, in between characters.

src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
jairaws: Yeah, I thought about that too. However, doing so fails at runtime, because pymxs can only be imported within 3dsMax. Since we want to use `MaxExecutableHandler` in the adaptor executable (e.g. `MaxAdaptor`), and not only the adaptor client (e.g. `DefaultMaxHandler`), that implementation throws an exception.

sakshie95 (approved):

src/deadline/max_adaptor/MaxAdaptor/regex_callback_handler.py
leongdl: Oh wow, in between characters. OK then this is the only way.

src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
leongdl: Got it, ok.

leongdl (approved):

PR #84, opened 2025-02-14
fix: Add length validation for QLineEdit widgets

jairaws:
The line input controls in the scene settings tab don't have a configured max length.

Set the max length of the line inputs to the max length allowed by OJD.

Addresses a security requirement.

* `hatch build`
*  `hatch run test`
*  Try to input a larger string than the max allowed in the submitter, and verified it was not allowed.

N/A

No

----


jairaws:
The Sonar analysis is failing due to some existing unit tests that are out of the scope of this change. The test folder structure was updated, and that's why those tests are being analyzed. The issue is that the tests are using a mock directory `"/tmp/9999"`, and this is triggering a low security risk due to the directory being global. However, the directory value is just a mock value used by the tests, it is not really accessed.

jericht (approved):

test/unit/deadline_submitter_for_3ds_max/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+    "pymxs",
+    "PySide2",
+    "PySide2.QtCore",
+    "PySide2.QtGui",
+    "PySide2.QtWidgets",
+    "qtmax",
+]
+
+for module in mock_modules:
+    sys.modules[module] = MagicMock()
```
leongdl, line 16: Never knew we could do this! Very cool.

leongdl (approved):

PR #86, opened 2025-02-20
fix: Add length validations for job settings input fields

jairaws:
Input controls in the job settings tab don't have length validations.

Update the sanity checks logic to validate the length of the inputs in the settings tab.

Improved security.

* `hatch build && hatch run fmt && hatch run all:test`
* Added unit tests.
* Tested end-to-end.

N/A.

No.

Screenshot 2025-02-19 at 3 54 38 PM

----


jairaws:
SonarCloud fails due to a risk with using. the random generator in the unit tests. It considers it to be weak cryptography. However, it is irrelevant because it is used only in the unit tests, and the intent is not related to cryptography.

src/deadline/max_submitter/sanity_checks.py
```diff
@@ -7,16 +7,23 @@
...
+from deadline.max_submitter.data_const import (
     ALL_CAMERAS_STR,
     ALL_STEREO_CAMERAS_STR,
     ALL_STATE_SETS_STR,
     ALLOWED_RENDERERS,
 )
 
 
+# This is the maximum string length allowed for OJD job parameters.
+JOB_PARAMETER_MAX_STRING_LENGTH: int = 1024
```
benl-2023, line 21: Do you know if we could fetch these constants from the python OJD model: https://github.com/OpenJobDescription/openjd-model-for-python?

May help us avoid keeping two definitions in sync over time.

src/deadline/max_submitter/sanity_checks.py
```diff
@@ -31,6 +38,26 @@ def check_sanity(settings: RenderSubmitterUISettings):
     # Check if any unsaved changes were made to the scene and prompt the user to save if not
     rt.checkForSave()
 
+    if len(settings.project_path) > JOB_PARAMETER_MAX_STRING_LENGTH:
```
benl-2023, line 41: Have we tested, and is it worth testing what happens if we put a non-path like value as the project path, ie setting project path to a random int?

jairaws: Didn't test it. The project path is read from the 3dsMax settings. The output path is selected from a browser widget. Furthermore, if for whatever reason the string is not a path, the backend should catch it. The validations done in the UI are meant to avoid more generic scenarios, like too big requests that could affect the service.

src/deadline/max_submitter/sanity_checks.py
jairaws: Yeah, that would be ideal. I don't know how to get the those limits from OJD, though. Definitely an improvement we can do later on.

benl-2023 (approved):

sakshie95 (approved):

PR #109, opened 2025-06-18
feat: Add submitter support for 3dsMax 2025+

### LINKED ISSUE #102: Bug: Submitter does not support 3ds Max 2025 / 2026
### Describe the bug

We would like support for 2025 and 2026.

I get this kind of error when I try on 2025.

-- Error occurred in anonymous codeblock; filename: C:\Program Files\Autodesk\3ds Max 2025\scripts\startup\STDCMenuCreator.ms; position: 654; line: 17
-- MAXScript Auto-load Script Error Exception:
-- Unknown property: "getMainMenuBar" in undefined
-- MAXScript callstack:
--	thread data: threadID:33264
--	------------------------------------------------------
--	[stack level: 0]
--	In anonymous codeblock; filename: C:\Program Files\Autodesk\3ds Max 2025\scripts\startup\STDCMenuCreator.ms; position: 655; line: 17
--		Locals:
--			debug: false
--			theDeadlineCloudMenu: undefined
--			theSubMenu: undefined
--			theMainMenu: undefined
--			theMenuDefinitionArray: undefined
--			menuMan: undefined
--		Externals:
--	------------------------------------------------------
--	[stack level: 1]
--	called from top-level

### Expected Behaviour

Be great to support 2025.

### Current Behaviour

Does not support 2025 or 2026

### Reproduction Steps

Try submitting from 2025

### Environment

At minimum:
1. Operating system (e.g. Windows Server 2022; Amazon Linux 2023; etc.)
2. Version of 3ds Max
3. Version of this package

Please share other details about your environment that you think might be relevant to reproducing the bug.

jairaws:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/issues/102

3dsMax 2025 has a couple of changes that affected the submitter: 1) The Max Script mechanism to handle menus changed, 2) Upgrade from PySide2 to PySide6.

* Create a new version of the menu with the new syntax.
* Remove direct references to PySide2.

Submitter support for 3dsMax 2025+

* Submitted job successfully from 3dsMax 2024 and 2025.
* Ran integ tests: `hatch run integ:test`
```
================================================= 2 passed in 56.38s ==================================================
```

Yes

No

Screenshot 2025-06-17 at 6 11 54 PM

----

### DIFF
--- a/install_files/STDCMenuCreator_v1.ms
+++ b/install_files/STDCMenuCreator_v1.ms
@@ -1,0 +1,55 @@
+-- Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+-- Compatibility: 3ds Max 2025+
+
+-- Function to create the AWS Deadline menu
+function createDeadlineCloudMenu =
+(
+    -- Get the menu manager from the callback parameter
+    local menuMgr = callbacks.notificationParam()
+    local mainMenuBar = menuMgr.mainMenuBar
+    
+    -- No need to check for existing menu - the CreateSubMenu will handle it
+    -- If a menu with the same GUID already exists, it will be replaced
+    
+    -- Create a new AWS Deadline menu with a unique GUID
+    -- Using a fixed GUID ensures consistency across sessions
+    local deadlineCloudMenuGUID = "F8FFB827-741C-4A81-8C89-BBF856DCF56D"
+    local deadlineCloudMenu = mainMenuBar.CreateSubMenu deadlineCloudMenuGUID "AWS Deadline"
+    
+    -- Define menu items with their GUIDs
+    local menuItems = #()
+    
+    -- Add the Submit to Deadline Cloud menu item
+    local submitActionGUID = "6AC5FDFE-2AF7-4069-8493-D8A1462F4E7F"
+    -- MacroScript action table ID is 647394
+    local macroScriptActionTableID = 647394
+    
+    -- Create the Submit to Deadline Cloud action
+    deadlineCloudMenu.CreateAction submitActionGUID macroScriptActionTableID "SubmitToDeadlineCloud`DeadlineCloud" title:"Submit to Deadline Cloud"
+    
+    -- Add the Test Job Bundle Output menu item if in DEBUG mode
+    local DEBUG = false
+    if DEBUG do
+    (
+        local testJobBundleGUID = "A2B3C4D5-E6F7-8901-2345-6789ABCDEF01"
+        deadlineCloudMenu.CreateAction testJobBundleGUID macroScriptActionTableID "TestJobBundleOutput`DeadlineCloud" title:"Run Job Bundle Tests"
+    )
+    
+    -- Python script to add repo path to max script locations
+    python.Execute "import sys"
+    python.Execute "import os"
+    python.Execute "sys.path.append(os.path.expanduser(r'~\\DeadlineCloudSubmitter\\Submitters\\3dsMax\\scripts'))"
+    
+    if DEBUG do
+    (
+        -- change to <reporoot>/src/deadline/max_submitter
+        -- TODO: Update to find local dev repo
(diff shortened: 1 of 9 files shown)


install_files/STDCMenuCreator_v1.ms
```diff
@@ -0,0 +1,55 @@
+-- Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+-- Compatibility: 3ds Max 2025+
```
jericht, line 2: Curious, is this just a note for humans? or does 3ds max recognize this and raise a warning if it tries to run it on a mismatched version?

pyproject.toml
```diff
@@ -31,7 +31,7 @@ classifiers = [
 # 2024 - 3.9: https://help.autodesk.com/view/MAXDEV/2024/ENU/?guid=MAXDEV_Python_about_the_3ds_max_python_api_html
 
 dependencies = [
-    "deadline == 0.49.*",
+    "deadline == 0.50.*",
```
jericht, line 34: Did you run the integ tests? Just want to be safe since there's a breaking change in this bump

install_files/STDCMenuCreator_v1.ms
```diff
@@ -0,0 +1,55 @@
...
+    local deadlineCloudMenuGUID = "F8FFB827-741C-4A81-8C89-BBF856DCF56D"
+    local deadlineCloudMenu = mainMenuBar.CreateSubMenu deadlineCloudMenuGUID "AWS Deadline"
+    
+    -- Define menu items with their GUIDs
+    local menuItems = #()
+    
+    -- Add the Submit to Deadline Cloud menu item
+    local submitActionGUID = "6AC5FDFE-2AF7-4069-8493-D8A1462F4E7F"
+    -- MacroScript action table ID is 647394
+    local macroScriptActionTableID = 647394
```
jericht, line 25: Is this a predefined constant in 3ds Max? or are we just choosing a static number here similar to the GUIDs?

install_files/STDCMenuCreator_v1.ms
```diff
@@ -0,0 +1,55 @@
...
+    
+    -- Create the Submit to Deadline Cloud action
+    deadlineCloudMenu.CreateAction submitActionGUID macroScriptActionTableID "SubmitToDeadlineCloud`DeadlineCloud" title:"Submit to Deadline Cloud"
+    
+    -- Add the Test Job Bundle Output menu item if in DEBUG mode
+    local DEBUG = false
```
jericht, line 31: nit: Could we put this in the top of the file with a comment similar to the v0? It looks like we need to manually set this to true to get these debug menus?

```
	-- change to true for development install
	local DEBUG = false
```

install_files/STDCMenuCreator_v1.ms
jairaws: Only a note for humans.

install_files/STDCMenuCreator_v1.ms
jairaws: This is based on the existing logic in the original version of this file. This debugging logic does not make much sense, but I don't want to change it right now. I'll work on another PR to automate the installation of the submitter, I'll see then how we can improve it.

install_files/STDCMenuCreator_v1.ms
jairaws: just a random value https://help.autodesk.com/view/MAXDEV/2025/ENU/?guid=GUID-FF48D0EC-6669-4EC7-AB43-E9998A14A198

leongdl (approved):
Question, where is the dependency for PyQt defined? Since we are changing the Qt import, would it cause issues?

leongdl:
Question, where is the dependency for PyQt defined? Since we are changing the Qt import, would it cause issues?

pyproject.toml
jairaws: Yes, I did. Updated the description to call it out.

benl-2023 (approved):

AlexTranAmz (approved):

PR #110, opened 2025-06-24
feat: Implement 3dsMax submitter installer

jairaws:
We need to automate the installation of the submitter.

* Create installer folder with all the components needed to define the submitter installer.
* Create scripts needed by the installer.
* Consolidate the menu in one single file for all versions.
* Unit tests for the installer.

Submitter installed automatically.

* Built submitter installer and installed submitter.
* Tested with 3ds Max 2024, 2025, 2026.
```
 hatch run installer:build-installer --install-builder-path "C:\\Program Files\\InstallBuilder Enterprise 25.3.1\\" --platform "windows" --local-dev
```

Yes

No

Screenshot 2025-06-23 at 8 27 41 PM

----


installer/DeadlineCloudFor3dsMaxSubmitter.xml
```diff
@@ -0,0 +1,244 @@
...
+                    </conditionRuleList>
+                    <actionList>
+                        <!-- InstallBuilder's system-wide Linux implementation does not work (it always writes to the user
+                        environment files like ~/.bashrc and ~/.bash_profile, even as root).
+                        Instead, we cache all environment variables in a global variable, and in the post-installation
+                        steps, we generate a file in /etc/profile.d/ that will get sourced on login. -->
+                        <globalVariables names="_fnAddEnvironmentVariable_list"/>
+                        <setInstallerVariable name="_fnAddEnvironmentVariable_list">
+                            <value>${_fnAddEnvironmentVariable_list}export ${name}=${value}&#xA;</value>
+                        </setInstallerVariable>
```
benl-2023: Curious, why do we need linux references in the installer if 3DSMax is only on windows?

scripts/build_installer.py
```diff
@@ -0,0 +1,181 @@
...
+from find_installbuilder import InstallBuilderSelection, get_builder_exe_name
+
+from deps_bundle import build_deps_bundle
+
+# This is derived from <installerFilename> in installer/DeadlineCloudClient.xml
+# See "Supported Platforms" table in https://releases.installbuilder.com/installbuilder/docs/installbuilder-userguide.html
+INSTALLER_FILENAMES = {
+    "Windows": "DeadlineCloudFor3dsMaxSubmitter-windows-x64-installer.exe",
+    "Linux": "DeadlineCloudFor3dsMaxSubmitter-linux-x64-installer.run",
+    "MacOS": "DeadlineCloudFor3dsMaxSubmitter-osx-installer.app",
```
benl-2023: Also curious why we have platforms other than windows here?

installer/DeadlineCloudFor3dsMaxSubmitter.xml
AlexTranAmz: Agree! We can remove this as it's only available on Windows

scripts/build_installer.py
AlexTranAmz: Same as above, I think for this case, since 3dsMax can only be installed on Windows, we only need a Windows installers

scripts/build_installer.py
```diff
@@ -0,0 +1,179 @@
...
+        )
+
+    if installer_platform == "Linux":
+        installbuilder_platform = "linux-x64"
+    elif installer_platform == "MacOS":
+        installbuilder_platform = "osx"
+    elif installer_platform == "Windows":
+        installbuilder_platform = "windows-x64"
+    else:
+        raise ValueError(f"Unknown platform '{installer_platform}'")
```
benl-2023, line 92: nit: probably don't need the non-windows platforms.

benl-2023 (approved):

scripts/build_installer.py
```diff
@@ -0,0 +1,179 @@
...
+        )
+        installer_dir = build_installer(
+            workdir=workdir,
+            component_file_path=installer_source_path,
+            install_builder_location=installbuilder_path,
+            dev=dev,
+            installer_platform=installer_platform,
+        )
+
+        installer_filename = INSTALLER_FILENAMES[installer_platform]
```
AlexTranAmz, line 160: qq: Since the `INSTALLER_FILENAMES` here could be empty if installer_platform is not Windows and thus create an error, could we change the installer platform to also only support windows and throw error of unknown platform otherwise?

AlexTranAmz (approved):

scripts/build_installer.py
AlexTranAmz: Sync offline with Jair! Agree that this is a minor improvement and not blocking this PR so I will approve it

PR #119, opened 2025-07-17
ci: add a workflow for manually releasing to pypi

moorec-aws:
We require a workflow for publishing to pypi in the event there is an issue publishing during the standard release process

Add a workflow dispatch workflow for manually releasing a tagged build to pypi

We can release to pypi if there is an issue doing so during the release process

Needs to be tested after merge

No

----

### DIFF
--- a/.github/workflows/manual_pypi_release.yml
+++ b/.github/workflows/manual_pypi_release.yml
@@ -1,0 +1,57 @@
+# Perform a manual PyPI release outside of the regular release process
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        required: true
+        default: ""
+        type: string
+
+jobs:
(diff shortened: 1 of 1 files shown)


leongdl (approved):

leongdl:
Thanks, seems like a copy paste from the other PRs in our other repos.

sakshie95:
qq. How are we planning to test this after merge? I would think not but confirming if there's any risks?

sakshie95 (approved):

moorec-aws:
> qq. How are we planning to test this after merge? I would think not but confirming if there's any risks?

Since the workflow publishes to pypi we cannot fully test the exact workflow before hand. We are copying our deployment from our release workflow here so there is no risk.

PR #132, opened 2025-08-21
Add Constants to support VRay7 GPU

leongdl:
- Users are requesting support of 3dsmax with VRay 7 GPU render. This case was not supported before.
- From max script, I listed the VRay7 render, the name for hot fix 2 were:
```
for r in rendererClass.classes do (
    format "Renderer: %\n" r
)
```
```
...
Renderer: V_Ray_GPU_7_Hotfix_2
Renderer: V_Ray_7_Hotfix_2
...
```

- Revisions:
 - Simplified the if-else VRayHandler logic. Did more testing.

- Lets add the missing constant, and set up the usage of the constant.

- 3dsmax with VRay 7 GPU support

- Built my own custom submitter, adapter.
- Created a job bundle, checked the bundle has the right renderer
- Used 3dsmax adapter via CLI and rendererd with VRay 7 GPU - https://github.com/leongdl/deadline-cloud-for-3ds-max/blob/testscript/test-3dsmax-openjd.ps1 
  - I made a prototype test script to convert the job bundle to openjd runner. This lets me test bundles as actual openjd 3dsmax adaptor commands.

- Not Applicable.

- No
- 
----

### DIFF
--- a/src/deadline/max_adaptor/MaxAdaptor/schemas/init_data.schema.json
+++ b/src/deadline/max_adaptor/MaxAdaptor/schemas/init_data.schema.json
@@ -37,7 +37,7 @@
             "oneOf": [
                 {
                     "type": "string",
-                    "pattern": "^V_Ray_([1-9]|GPU_[1-9]).*$"
+                    "pattern": "^V_Ray.*$"
                 },
                 {
                     "enum": [
--- a/src/deadline/max_adaptor/MaxClient/render_handlers/__init__.py
+++ b/src/deadline/max_adaptor/MaxClient/render_handlers/__init__.py
@@ -23,11 +23,9 @@ def get_render_handler(renderer: str = "Default_Scanline_Renderer") -> DefaultMa
         return ArtHandler()
     elif renderer == "Corona":
         return CoronaHandler()
-    elif renderer.startswith("V_Ray_6"):
-        return VrayHandler(gpu=False)
-    elif renderer.startswith("V_Ray_GPU_6"):
+    elif renderer.startswith("V_Ray_GPU_"):
         return VrayHandler(gpu=True)
-    elif renderer.startswith("V_Ray_7"):
+    elif renderer.startswith("V_Ray_"):
         return VrayHandler(gpu=False)
     elif renderer == "Redshift_Renderer":
         return RedshiftHandler()
--- a/src/deadline/max_submitter/sanity_checks.py
+++ b/src/deadline/max_submitter/sanity_checks.py
@@ -177,10 +177,16 @@ def check_sanity_specific_state_set(settings: RenderSubmitterUISettings, state_s
             f"The state set name {state_set} is too long. The max length allowed is {STEP_NAME_MAX_STRING_LENGTH}."
         )
 
-    if str(rt.renderers.current).split(":")[0].split("__")[0] not in ALLOWED_RENDERERS:
+    renderer_name = str(rt.renderers.current).split(":")[0]
+    renderer_split_underscore = renderer_name.split("__")[0]
+    renderer_split_single_underscore = "_".join(renderer_name.split("_")[:4])
+
+    if (
+        renderer_split_underscore not in ALLOWED_RENDERERS
+        and renderer_split_single_underscore not in ALLOWED_RENDERERS
+    ):
         raise Exception(
-            f"{state_set} has an unsupported renderer set. Renderer: "
-            f"{str(rt.renderers.current).split(':')[0]}"
+            f"{state_set} has an unsupported renderer set. Renderer: " f"{renderer_name}"
         )
 
     if not settings.override_frame_range:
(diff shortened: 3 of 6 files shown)


src/deadline/max_adaptor/MaxClient/render_handlers/__init__.py
```diff
@@ -29,6 +29,8 @@ def get_render_handler(renderer: str = "Default_Scanline_Renderer") -> DefaultMa
         return VrayHandler(gpu=True)
     elif renderer.startswith("V_Ray_7"):
         return VrayHandler(gpu=False)
+    elif renderer.startswith("V_Ray_GPU_7"):
```
lucaseck: Can the versions be removed here. e.g. renderer.startswith("V_Ray_GPU_") and renderer.startswith("V_Ray_") after? One less place to update when adding support for new versions I think.

leongdl: Sure, easy todo.

leongdl: Updated - GPU goes first since the matching order will affect the logic now.

src/deadline/max_adaptor/MaxAdaptor/schemas/init_data.schema.json
```diff
@@ -37,7 +37,7 @@
             "oneOf": [
                 {
                     "type": "string",
-                    "pattern": "^V_Ray_([1-9]|GPU_[1-9]).*$"
+                    "pattern": "^V_Ray.*$"
```
leongdl, line 40: We have the other checks in place, this is too strict and not useful in the adaptor schema

src/deadline/max_adaptor/MaxClient/render_handlers/__init__.py
```diff
@@ -23,11 +23,9 @@ def get_render_handler(renderer: str = "Default_Scanline_Renderer") -> DefaultMa
         return ArtHandler()
     elif renderer == "Corona":
         return CoronaHandler()
-    elif renderer.startswith("V_Ray_6"):
-        return VrayHandler(gpu=False)
-    elif renderer.startswith("V_Ray_GPU_6"):
+    elif renderer.startswith("V_Ray_GPU_"):
         return VrayHandler(gpu=True)
-    elif renderer.startswith("V_Ray_7"):
+    elif renderer.startswith("V_Ray_"):
```
leongdl, line 28: Moved GPU first, since the if-elif blocks now affect the logic without the VRay version. (As suggested by Lucas)

src/deadline/max_submitter/sanity_checks.py
```diff
@@ -177,10 +177,14 @@ def check_sanity_specific_state_set(settings: RenderSubmitterUISettings, state_s
             f"The state set name {state_set} is too long. The max length allowed is {STEP_NAME_MAX_STRING_LENGTH}."
         )
 
-    if str(rt.renderers.current).split(":")[0].split("__")[0] not in ALLOWED_RENDERERS:
+    renderer_name = str(rt.renderers.current).split(":")[0]
+    renderer_split_underscore = renderer_name.split("__")[0]
+    renderer_split_single_underscore = "_".join(renderer_name.split("_")[:4])
+
```
leongdl: Unit tests added below to test this.

src/deadline/max_adaptor/MaxClient/render_handlers/__init__.py
godobyte: Do we have any checks prior to calling this function, could removing V_Ray version lead to handler being created for an unsupported renderer? What happens in that case, i.e. returning VrayHandler rather than DefaultMaxHandler?

leongdl: No - that will not happen, because the `VRay_` and `VRay_GPU` are the nomenclature of the VRay plugin. 

That's the "standard",

src/deadline/max_submitter/sanity_checks.py
```diff
@@ -177,10 +177,16 @@ def check_sanity_specific_state_set(settings: RenderSubmitterUISettings, state_s
             f"The state set name {state_set} is too long. The max length allowed is {STEP_NAME_MAX_STRING_LENGTH}."
         )
 
-    if str(rt.renderers.current).split(":")[0].split("__")[0] not in ALLOWED_RENDERERS:
+    renderer_name = str(rt.renderers.current).split(":")[0]
+    renderer_split_underscore = renderer_name.split("__")[0]
+    renderer_split_single_underscore = "_".join(renderer_name.split("_")[:4])
```
jericht, line 182: nit: the unit test cases make this string splitting a lot easier to understand! Can we put an example value of the full `rt.renderers.current` string here as a comment for readability? like `V_Ray_6__update_2_1:V-Ray 6, update 2.1`

jericht (approved):

src/deadline/max_adaptor/MaxClient/render_handlers/__init__.py
godobyte: Synced offline, I meant standard case like like `V_Ray_5`, this function will return VrayHandler rather than DefaultMaxHandler, which is better since DefaultMaxHandler is for Arnold.

godobyte (approved):

PR #136, opened 2025-08-26
fix: use correct VRay regex in adaptor init-data schema

muratmozdemir:
While running 3DSMax on Windows using VRay6 GPU (Service-Managed Fleet), we get this error from jobs:
```
Error encountered while starting adaptor: 'V_Ray_GPU_6__update_2_1' 
does not match '^V_Ray_[1-9](_GPU)?.*$'
```

If I use PowerShell to change the regex string to `"^V_Ray_([1-9]|GPU_[1-9]).*$"`, this resolves the issue. (This is done with a host configuration script). This way it doesn't matter if its `V_Ray_GPU_6__update_2_1` or `V_Ray_6_GPU__update_2_1`, however, if you look at [this](https://github.com/aws-deadline/deadline-cloud-for-3ds-max/blob/8cea88ec48848c4f0324e5e63ba7d09dca5a2ac6/src/deadline/max_submitter/data_const.py#L27) I believe `V_Ray_GPU_6` is the correct string.

This should resolve issues with VRay6 GPU renderers on 3DSMax

I have used PowerShell and host configuration scripts to modify the schema directly on a server before a job start which resolved this issue

Currently, yes.  But the new regex should handle both strings. (`V_Ray_GPU_6__update_2_1` or `V_Ray_6_GPU__update_2_1`)

### DIFF
--- a/src/deadline/max_adaptor/MaxAdaptor/schemas/init_data.schema.json
+++ b/src/deadline/max_adaptor/MaxAdaptor/schemas/init_data.schema.json
@@ -37,7 +37,7 @@
             "oneOf": [
                 {
                     "type": "string",
-                    "pattern": "^V_Ray_[1-9](_GPU)?.*$"
+                    "pattern": "^V_Ray_([1-9]|GPU_[1-9]).*$"
                 },
                 {
                     "enum": [


jericht:
Thanks for contributing @muratmozdemir! These changes look good to me. You'll need to update your commit message to match our [configured conventional commits tags](https://github.com/aws-deadline/deadline-cloud-for-3ds-max/blob/0f573a7193325f0e21ed1233a08dfc52f6af9d8a/pyproject.toml#L160-L171) (change it to `fix:`) and also sign your commit please! Happy to approve once those are done.


You can run something like this
```
git commit --amend -s -m "fix: use correct VRay regex in adaptor init-data schema"
```

muratmozdemir:
> Thanks for contributing @muratmozdemir! These changes look good to me. You'll need to update your commit message to match our [configured conventional commits tags](https://github.com/aws-deadline/deadline-cloud-for-3ds-max/blob/0f573a7193325f0e21ed1233a08dfc52f6af9d8a/pyproject.toml#L160-L171) (change it to `fix:`) and also sign your commit please! Happy to approve once those are done.
> 
> You can run something like this
> 
> ```
> git commit --amend -s -m "fix: use correct VRay regex in adaptor init-data schema"
> ```

This should all be fixed now. 

Thanks!!

jericht (approved):

leongdl (approved):

PR #138, opened 2025-08-28
doc: update development.md with installer build instructions 

leongdl:
- It was hard to setup 3dsmax submitter via the manual copy instructions.

- Now that we support install builder, it is a much better development experience to update and test the submitter.

- Better dev flow.

- I tried this way of development and updating the submitter!

- This is a doc change

- NA.

----


AlexTranAmz (approved):

jericht (approved):
🚢

PR #149, opened 2025-09-12
fix: Fix V_Ray renderer generally using "starts with"

leongdl:
- I found another path where the VRay renderer detection was problematic.
- It seems VRay is inconsistency using _ or __ and different formatting for the renderer name. This is quite troublesome.

- At least the V_Ray begins with string is always consistent and more passible. 
- Use Begins with instead, and add unit test to detect the cases.

- Better VRay detection in 3dsmax

- Loaded a test scene, selected different VRay7 renderers, GPU, CPU, ART etc. Started the Deadline Cloud submitter to export a bundle. This will trigger the submitter sanity test code.
 
- No

- No

----

### DIFF
--- a/src/deadline/max_submitter/sanity_checks.py
+++ b/src/deadline/max_submitter/sanity_checks.py
@@ -178,13 +178,17 @@ def check_sanity_specific_state_set(settings: RenderSubmitterUISettings, state_s
         )
 
     renderer_name = str(rt.renderers.current).split(":")[0]
-    renderer_split_underscore = renderer_name.split("__")[0]
-    renderer_split_single_underscore = "_".join(renderer_name.split("_")[:4])
 
-    if (
-        renderer_split_underscore not in ALLOWED_RENDERERS
-        and renderer_split_single_underscore not in ALLOWED_RENDERERS
-    ):
+    # Check if renderer is supported - either exact match or starts with an allowed renderer
+    renderer_supported = renderer_name in ALLOWED_RENDERERS
+    if not renderer_supported:
+        # Check if renderer starts with any allowed renderer (handles versions/hotfixes)
+        for allowed_renderer in ALLOWED_RENDERERS:
+            if renderer_name.startswith(allowed_renderer):
+                renderer_supported = True
+                break
+
+    if not renderer_supported:
         raise Exception(
             f"{state_set} has an unsupported renderer set. Renderer: " f"{renderer_name}"
         )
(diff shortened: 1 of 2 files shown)


src/deadline/max_submitter/sanity_checks.py
```diff
@@ -178,13 +178,17 @@ def check_sanity_specific_state_set(settings: RenderSubmitterUISettings, state_s
...
-    if (
-        renderer_split_underscore not in ALLOWED_RENDERERS
-        and renderer_split_single_underscore not in ALLOWED_RENDERERS
-    ):
+    # Check if renderer is supported - either exact match or starts with an allowed renderer
+    renderer_supported = renderer_name in ALLOWED_RENDERERS
+    if not renderer_supported:
+        # Check if renderer starts with any allowed renderer (handles versions/hotfixes)
+        for allowed_renderer in ALLOWED_RENDERERS:
+            if renderer_name.startswith(allowed_renderer):
```
jairaws, line 187: This approach would be a problem if any of the allowed renderers is a substring of one of the other allowed renderers. I see that that is not a problem right now.
https://github.com/aws-deadline/deadline-cloud-for-3ds-max/blob/mainline/src/deadline/max_submitter/data_const.py#L23

Could we add a unit test to make sure that none of the allowed renders is a substring of the others? This would prevent failing into this problem in the future.

leongdl: Interesting question, yes we absolutely can add that test. Let me write it!

leongdl: Implemented!

test/unit/deadline_submitter_for_3ds_max/test_sanity_checks.py
```diff
@@ -191,3 +203,25 @@ def test_check_sanity_specific_state_set_unsupported_renderer(
...
         with raises(Exception, match="has an unsupported renderer set"):
             check_sanity_specific_state_set(default_settings, "test_state_set")
+
+    def test_allowed_renderers_no_substring_conflicts(self) -> None:
+        """Test that none of the allowed renderers is a substring of another.
+
+        This prevents issues where renderer version checking might incorrectly match
+        a shorter renderer name that is a substring of a longer one.
+        For example, if 'V_Ray' and 'V_Ray_6' were both in the list,
+        'V_Ray_6_update_1' would match 'V_Ray' first instead of 'V_Ray_6'.
```
benl-2023, line 213: Would we ever have just "V_Ray" in the ALLOWED_RENDERERS list?

benl-2023 (approved):

jairaws (approved):

PR #160, opened 2025-10-19
feat: Add new shared utilities class to support render element and ge…

leongdl:
### REVIEWER:
- Please read - https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/153 for a full code level walk through of this PR.
- This review is a chunked subset of changes specifically for the shared utility changes from https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/152. I did this to make smaller reviews for the reviewer.

### What was the problem/requirement? (What/Why)

The 3ds Max integration for Deadline Cloud lacked comprehensive support for render elements, which are essential for advanced rendering workflows in 3ds Max. Users needed the ability to configure, validate, and manage render elements (such as diffuse, reflection, shadow passes, etc.) through both the submitter UI and the adaptor during job execution. This functionality was critical for professional rendering pipelines that rely on multi-pass rendering and compositing workflows.

### What was the solution? (How)

The solution implemented comprehensive render elements support across the entire 3ds Max integration:

- Created a new RenderElementManager class in the adaptor to handle render element configuration during job execution
- Added shared utilities (max_shared/utilities/max_utils.py) containing 829+ lines of common render element functionality used by both submitter and adaptor
- Implemented a new render elements UI widget (render_elements_widget.py) with 467+ lines for user configuration in the submitter
- Extended job bundle generation to include render element parameters and validation
- Updated schemas and templates to support render element data in the job workflow
- Added comprehensive validation and sanity checks for render element configurations
- Integrated V-Ray VFB support and split buffer functionality for V-Ray renderer compatibility

### What is the impact of this change?

This change enables professional 3ds Max users to leverage advanced multi-pass rendering workflows through Deadline Cloud, significantly expanding the integration's capabilities. Users can now:
- Configure render elements through the submitter UI
- Have render elements automatically managed during cloud rendering
- Support complex V-Ray rendering setups with VFB integration
- Maintain consistent render element behavior between local and cloud rendering

### How was this change tested?

The change includes comprehensive unit tests for the shared utilities (test_shared_max_utils.py) and extends existing adaptor tests (test_max_handler_base.py). The implementation follows established patterns from Deadline 10's render element system, ensuring compatibility and reliability.

### Was this change documented?

Yes, the change includes a new README file (src/deadline/max_shared/README.md) documenting the shared utilities and their usage patterns. The code is extensively commented with docstrings explaining the render element functionality.

### Is this a breaking change?

[...]

PR #153, opened 2025-09-17
doc: render elements implementation document

leongdl:
### What was the problem/requirement? (What/Why)
AWS Deadline Cloud for 3ds Max had only basic render elements support, lacking the comprehensive functionality that existed in Deadline 10. Users needed advanced render element features including path management with naming patterns, V-Ray VFB integration, element ignore functionality, and automatic configuration during rendering. The existing implementation was limited to basic enable/disable functionality without the sophisticated workflow management that production environments require.

To read more about Render Elements in 3dsMax, please refer to Autodesk documentation [here](https://docs.chaos.com/display/VMAX/Render+Elements).

### What was the solution? (How)
Implemented comprehensive render elements support through a three-tier architecture: (1) Enhanced submitter with 8 new render element properties in RenderSubmitterUISettings and corresponding UI components, (2) Shared utilities (max_shared/utilities/max_utils.py) providing consistent pymxs operations between submitter and adaptor, (3) Full adaptor integration with RenderElementManager class and automatic workflow integration in all render handlers. The solution includes parameter flow from UI settings through OpenJD job templates to adaptor execution, with proper state backup and restoration.

### What is the impact of this change?
Users can now configure render elements with the same level of control as Deadline 10, including advanced path management, V-Ray VFB control, split buffer support, and element ignore functionality. The implementation provides automatic render element configuration before each frame render and proper cleanup afterward, ensuring consistent behavior across all supported renderers (Default, V-Ray, Corona, ART, Redshift). This brings Deadline Cloud for 3ds Max to feature parity with Deadline 10 for render elements functionality.

### How was this change tested?
The implementation was tested through direct adaptor testing using PowerShell scripts with template parameter substitution, validating complete parameter flow from UI through job templates to 3ds Max scene configuration. Testing included verification of pymxs operations, render element detection, path management, V-Ray integration, and state restoration. The shared utilities architecture ensures consistent behavior between submitter and adaptor components.

### Was this change documented?
Yes, comprehensive documentation was created including a detailed design document (render-element-design.md) covering system architecture, data structures, implementation details, and pymxs operations mapping. The documentation includes block diagrams, sequence diagrams, and complete technical specifications for all components. Additionally, a reusable pattern document (make-design-doc.md) was created for future feature implementations.

### Is this a breaking change?
[...]

PR #161, opened 2025-10-20
feat(render elements) Add Submitter UI to support Render Elements

leongdl:
### REVIEWER:
- Please read - https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/153 for a full code level walk through of this PR.
- This review is a chunked subset of changes specifically for the submitter UI. This is PR 2 of 3 for the render elements feature.

PR1 - Share utils: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/160
PR2 - Submission UI: this PR.
PR3 - Adapter integration. https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/162

The 3ds Max integration for Deadline Cloud lacked comprehensive support for render elements, which are essential for advanced rendering workflows in 3ds Max. Users needed the ability to configure, validate, and manage render elements (such as diffuse, reflection, shadow passes, etc.) through both the submitter UI and the adaptor during job execution. This functionality was critical for professional rendering pipelines that rely on multi-pass rendering and compositing workflows.

- In this PR, I am adding the UI elements in the 3dsmax submitter dialog. 
- The dialog has an on/off toggle (default off) to allow users to modify render elements. 
  - When off, the scene renders as is, and does not modify any render element properties. When 
  - When on, the scene applies the selected render elements on / off / properties.
- When clicking export bundle or render, the corresponding job template parameters are generated for submission.

This change enables professional 3ds Max users to leverage advanced multi-pass rendering workflows through Deadline Cloud, significantly expanding the integration's capabilities. Users can now:
- Configure render elements through the submitter UI
- Have render elements automatically managed during cloud rendering
- Support complex V-Ray rendering setups with VFB integration
- Maintain consistent render element behavior between local and cloud rendering

### Specific UI testing:
Screenshot 2025-10-20 at 4 44 37 PM
Screenshot 2025-10-20 at 4 44 28 PM

- Tested both positive and negative cases for all buttons. 
- When disabled, the widget greys out all options. This can be a simple toggle to turn on the feature.
- When enabled, the widget enables all options. The settings are sticky too.
- Tested Job Bundle parameters both on/off combinations to make sure bundles are generated correctly.

- We will create a "user guide" in the future, but this is a UX change.

No, this is not a breaking change. The render elements support is additive functionality that extends existing capabilities without modifying existing APIs or workflows.

----

### DIFF
--- a/src/deadline/max_submitter/create_job_bundle.py
+++ b/src/deadline/max_submitter/create_job_bundle.py
@@ -4,16 +4,25 @@
 3ds Max Deadline Cloud Submitter - Functions for generating the job template and parameter values files
 """
 
+import logging
 import os
 from copy import deepcopy
 from pathlib import Path
 from typing import Any
 
 import yaml
-from data_classes import RenderSubmitterUISettings, StateSetData
+from data_classes import (
+    RENDER_ELEMENT_PARAMS,
+    RENDER_ELEMENT_PARAM_MAPPING,
+    RenderSubmitterUISettings,
+    StateSetData,
+)
 from data_const import ALL_CAMERAS_STR, ALL_STATE_SETS_STR, ALL_STEREO_CAMERAS_STR
 from deadline.client.exceptions import DeadlineOperationError
 from utilities import max_utils
+from deadline.max_shared.utilities.max_utils import get_render_elements
+
+_logger = logging.getLogger(__name__)
 
 
 def get_job_template(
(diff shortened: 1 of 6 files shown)


src/deadline/max_submitter/create_job_bundle.py
```diff
@@ -107,6 +113,171 @@ def _create_param_definitions(
             }
         )
 
+    # Add render elements parameters if render elements are present in the scene
+    render_elements = get_render_elements()
+    if render_elements:
+        # RenderElementsModified parameter - whether render elements settings are being modified
```
leongdl: I'm going to rephrase this to `# Enabled to modify RenderElement settings at render time.`

src/deadline/max_submitter/create_job_bundle.py
```diff
@@ -50,6 +56,180 @@ def get_job_template(
     return job_template
 
 
+def create_job_bundle_render_element_props(
```
leongdl: Comment from Jair before:
> This logic should be factored out from this method. _create_param_definitions() is already too long.

I have updated and split this out to a separate method.

src/deadline/max_submitter/create_job_bundle.py
```diff
@@ -358,6 +551,91 @@ def _get_job_parameters(
     ):
         parameter_values.append({"name": "Camera", "value": settings.camera_selection})
 
+    # Add render elements parameters if render elements are present in the scene
+    render_elements = get_render_elements()
+    if render_elements:
+        # Enabled to modify RenderElement settings at render time.
+        parameter_values.append(
```
leongdl, line 561: Question from Jair
> So much code repetition. We should refactor this code.

I tried to refactor this as individual methods, or a for loop, but since each property can be a boolean or a string list. Couldn't find a good compromise so I'll keep it as a `parameter_value.append`....

src/deadline/max_submitter/ui/render_elements_widget.py
```diff
@@ -0,0 +1,531 @@
...
+        - ❌ (Cross mark) = Error indicator
+
+        Args:
+            text (str): Text that may contain status emoji characters
+
+        Returns:
+            str: Text with status emojis removed and whitespace cleaned up
+        """
+        # Remove specific status emojis used in this widget
+        clean_text = text.replace("🟢", "").replace("🟡", "").replace("🔴", "").replace("❌", "")
```
benl-2023, line 71: Is there a range of unicode characters we could strip more comprehensively? Or does green, yellow, red, X suffice for this use case?

src/deadline/max_submitter/ui/render_elements_widget.py
```diff
@@ -0,0 +1,531 @@
...
+        main_layout.addWidget(render_elements_group)
+
+        # Row 0: Modify Render Elements checkbox (master control)
+        self.modify_render_elements_checkbox = QCheckBox("Modify Render Elements")
+        self.modify_render_elements_checkbox.setChecked(False)  # Default unchecked
+        self.modify_render_elements_checkbox.setToolTip(
+            "Enable or disable modification of render elements settings. "
+            "When unchecked, all render elements setting changes on the worker will be disabled."
+        )
+        layout.addWidget(self.modify_render_elements_checkbox, 0, 0, 1, 2)
```
benl-2023, line 94: I see a lot of magic numbers here, does it make sense to define them as constants?

Alternatively, would using keyword args make this more readable?

src/deadline/max_submitter/ui/render_elements_widget.py
```diff
@@ -0,0 +1,531 @@
...
+            for element in render_elements:
+                name = element.name
+                element_type = element.type
+                enabled = element.enabled
+                has_output = element.has_output_path
+                output_filename = element.output_filename
+
+                # Status indicators
+                status_parts = []
+                status_icon = "🟢"  # Default green for enabled with output
```
benl-2023, line 371: Ah, I see this is where the emojis come from.

benl-2023: Would the implementation be cleaner if we handled the display text (with emojis) separate from the element names?

src/deadline/max_submitter/ui/scene_settings_tab.py
```diff
@@ -232,13 +233,23 @@ def _build_ui(self, settings):
...
 
+        # Render elements widget
+        self.render_elements_widget = RenderElementsWidget(settings, self)
+        self.render_elements_widget.settings_changed.connect(
+            self._on_render_elements_settings_changed
+        )
+        self.render_elements_widget.validation_changed.connect(
+            self._on_render_elements_validation_changed
+        )
+        lyt.addWidget(self.render_elements_widget, 12, 0, 4, 5)
```
benl-2023, line 241: Same comment about magic numbers

src/deadline/max_submitter/data_classes.py
```diff
@@ -8,13 +8,41 @@
...
+    "RenderElementsIncludeNameInPath": "render_elements_include_name_in_path",
+    "RenderElementsIncludeTypeInPath": "render_elements_include_type_in_path",
+    "RenderElementsIncludeNameInFilename": "render_elements_include_name_in_filename",
+    "RenderElementsIncludeTypeInFilename": "render_elements_include_type_in_filename",
+    "VRayRenderElementsVFBControl": "vray_render_elements_vfb_control",
+    "VRaySplitBufferSupport": "vray_split_buffer_support",
+    "IgnoreRenderElementsByName": "ignore_render_elements_by_name",
+}
+
+
```
benl-2023, line 45: Would this be better defined as a String Enum?

leongdl: Nope, because this is a string map, and we use it as a map in the file above. 

This is just a short hand for a Snake case to Camel case map to save on computing translating it.

src/deadline/max_submitter/ui/scene_settings_tab.py
leongdl: Can't change this since  these are layout guidelines and cannot be hard coded.

leongdl: See my notes above.

src/deadline/max_submitter/data_classes.py
```diff
@@ -78,6 +106,41 @@ class RenderSubmitterUISettings:
     all_cameras: list[str] = field(default_factory=list)
     all_stereo_cameras: list[str] = field(default_factory=list)
 
+    # Render Elements (Basic Support - already implemented)
+    # Master control for render elements modification
```
jericht, line 110:
```suggestion
    # Main control for render elements modification
```

src/deadline/max_submitter/ui/scene_settings_tab.py
```diff
@@ -232,13 +233,23 @@ def _build_ui(self, settings):
         self._build_scene_tweaks_ui()
         lyt.addWidget(self.scene_tweaks_grp_box, 9, 0, 3, 5)
```
leongdl, line 234: The layout of the widget is hard coded all across this tab, https://doc.qt.io/qt-6/qgridlayout.html, these are row, column, and span layouts - https://doc.qt.io/qt-6/qgridlayout.html

src/deadline/max_submitter/ui/scene_settings_tab.py
leongdl: The layout of the widget is hard coded all across this tab, https://doc.qt.io/qt-6/qgridlayout.html, these are row, column, and span layouts - https://doc.qt.io/qt-6/qgridlayout.html

src/deadline/max_submitter/ui/render_elements_widget.py
leongdl: These icons are part of this PR and created by copying D10, these 4 icons are the exhaustive list.

src/deadline/max_submitter/create_job_bundle.py
```diff
@@ -50,6 +56,180 @@ def get_job_template(
...
+    """
+    Creates render element parameter definitions and adds them to the job template.
+
+    :param job_template: the job template to modify
+    :param settings: a RenderSubmitterUISettings object containing the latest UI settings
+    """
+    render_elements = get_render_elements()
+    if render_elements:
+        # Enabled to modify RenderElement settings at render time.
+        job_template["parameterDefinitions"].append(
```
jericht, line 74: What if `job_template` does not have `parameterDefinitions`? or it is not a list?

src/deadline/max_submitter/ui/render_elements_widget.py
leongdl: Unfortunately not, we aren't using the nice Qt GUI layout system that does not use magic numbers. Our entire repo of UI layouts uses magic numbers.

leongdl: If only we used something like this: https://doc.qt.io/qt-6/designer-using-a-ui-file.html

src/deadline/max_submitter/ui/render_elements_widget.py
leongdl: This was something ported from D10, as a signal what render elements work or not.

benl-2023 (dismissed):

src/deadline/max_submitter/create_job_bundle.py
```diff
@@ -403,6 +681,43 @@ def _get_queue_parameters_for_bundle(
...
+            if invalid_names:
+                raise DeadlineOperationError(
+                    f"The following render element names to ignore do not exist in the scene: "
+                    f"{', '.join(invalid_names)}"
+                )
+        except Exception as e:
+            # If validation fails, log warning but don't fail submission
+            import logging
+
+            _logger = logging.getLogger(__name__)
```
jericht: nit: how about adding this to the module level?

src/deadline/max_submitter/ui/scene_settings_tab.py
```diff
@@ -282,6 +293,30 @@ def _build_scene_tweaks_ui(self):
...
+    def _on_render_elements_settings_changed(self):
+        """
+        Handle changes from the enhanced render elements widget.
+
+        The render elements widget manages its own settings internally,
+        so we just need to acknowledge the change occurred.
+        """
+        _logger.debug("Render elements settings changed")
+        # The render elements widget handles its own settings updates
+        # No additional action needed here
```
jericht: Do we need this handler if we're not doing anything in it? Can we remove this?

src/deadline/max_submitter/ui/render_elements_widget.py
```diff
@@ -0,0 +1,531 @@
...
+                self.detected_elements_list.addItem(item)
+                return
+
+            # Display render elements with enhanced information
+            for element in render_elements:
+                name = element.name
+                element_type = element.type
+                enabled = element.enabled
+                has_output = element.has_output_path
+                output_filename = element.output_filename
```
jericht, line 367: nit: do we need these variables? Looks like we're just using these to fill in text strings below

src/deadline/max_submitter/ui/render_elements_widget.py
```diff
@@ -0,0 +1,531 @@
...
+        # If we're enabling, also respect the render elements checkbox state
+        if enabled:
+            self._on_render_elements_changed(self.render_elements_checkbox.checkState())
+
+        self._on_settings_changed()
+
+    def _on_render_elements_changed(self, state):
+        """
+        Handle changes to the main render elements checkbox.
+        Only affects controls if the master modify checkbox is enabled.
```
jericht:
```suggestion
        Only affects controls if the main modify checkbox is enabled.
```

src/deadline/max_submitter/ui/render_elements_widget.py
```diff
@@ -0,0 +1,531 @@
...
+        """
+        # Only enable/disable controls if the master modify checkbox is checked
```
jericht:
```suggestion
        # Only enable/disable controls if the main modify checkbox is checked
```

src/deadline/max_submitter/create_job_bundle.py
leongdl: This is a deep copy of the default job template, so it will be OK - https://github.com/aws-deadline/deadline-cloud-for-3ds-max/blob/cc2a1fe5f15f696f884a0461ed9c7fba0a3e8514/src/deadline/max_submitter/default_max_job_template.yaml

leongdl: And also similar pattern to the other code.

src/deadline/max_submitter/create_job_bundle.py
leongdl: Good point, this is an AI BUG!!

leongdl: Done!

src/deadline/max_submitter/ui/scene_settings_tab.py
leongdl: Deleted!

src/deadline/max_submitter/ui/render_elements_widget.py
leongdl: Just a stylistic formatting only

src/deadline/max_submitter/ui/render_elements_widget.py
leongdl: Thanks! Good catch on word usage

jericht (approved):

benl-2023 (approved):

PR #162, opened 2025-10-22
feat(render elements): Adapter support for render elements. Apply settings from job template via pyxms

leongdl:
### REVIEWER:
- Please read - https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/153 for a full code level walk through of this PR.
- This review is a chunked subset of changes specifically for the submitter UI. This is PR 2 of 3 for the render elements feature.

PR1 - Share utils: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/160
PR2 - Submission UI: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/161
PR3 - Adapter https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/162

### What was the problem/requirement? (What/Why)

The 3ds Max integration for Deadline Cloud lacked comprehensive support for render elements, which are essential for advanced rendering workflows in 3ds Max. Users needed the ability to configure, validate, and manage render elements (such as diffuse, reflection, shadow passes, etc.) through both the submitter UI and the adaptor during job execution. This functionality was critical for professional rendering pipelines that rely on multi-pass rendering and compositing workflows.

### What was the solution? (How)
- In this PR, I am adding the adapter changes required to process render element changes and apply them prior to rendering the scene. 
- The adapter takes the parameters from the job bundle, changes each setting at render time.

### What is the impact of this change?

This change enables professional 3ds Max users to leverage advanced multi-pass rendering workflows through Deadline Cloud, significantly expanding the integration's capabilities. Users can now:
- Configure render elements through the submitter UI
- Have render elements automatically managed during cloud rendering
- Support complex V-Ray rendering setups with VFB integration
- Maintain consistent render element behavior between local and cloud rendering

### How was this change tested?
- Render testing, some results will be added to the PR.

### Was this change documented?
- We will create a "user guide" in the future, but this is a UX change.

### Is this a breaking change?

No, this is not a breaking change. The render elements support is additive functionality that extends existing capabilities without modifying existing APIs or workflows.

----
*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

PR #152, opened 2025-09-17
feat: add render elements support

leongdl:
### REVIEWER:
- Please read - https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/153 for a full code level walk through of this PR.

- This PR has been cut up into 3 smaller PRs for easier reviews.
PR1 - Share utils: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/160
PR2 - UI dialog and job template generation: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/161
PR3 - Adapter integration. WIP.

### What was the problem/requirement? (What/Why)

The 3ds Max integration for Deadline Cloud lacked comprehensive support for render elements, which are essential for advanced rendering workflows in 3ds Max. Users needed the ability to configure, validate, and manage render elements (such as diffuse, reflection, shadow passes, etc.) through both the submitter UI and the adaptor during job execution. This functionality was critical for professional rendering pipelines that rely on multi-pass rendering and compositing workflows.

### What was the solution? (How)

The solution implemented comprehensive render elements support across the entire 3ds Max integration:

- Created a new RenderElementManager class in the adaptor to handle render element configuration during job execution
- Added shared utilities (max_shared/utilities/max_utils.py) containing 829+ lines of common render element functionality used by both submitter and adaptor
- Implemented a new render elements UI widget (render_elements_widget.py) with 467+ lines for user configuration in the submitter
- Extended job bundle generation to include render element parameters and validation
- Updated schemas and templates to support render element data in the job workflow
- Added comprehensive validation and sanity checks for render element configurations
- Integrated V-Ray VFB support and split buffer functionality for V-Ray renderer compatibility

### What is the impact of this change?

This change enables professional 3ds Max users to leverage advanced multi-pass rendering workflows through Deadline Cloud, significantly expanding the integration's capabilities. Users can now:
- Configure render elements through the submitter UI
- Have render elements automatically managed during cloud rendering
- Support complex V-Ray rendering setups with VFB integration
- Maintain consistent render element behavior between local and cloud rendering

### How was this change tested?

The change includes comprehensive unit tests for the shared utilities (test_shared_max_utils.py) and extends existing adaptor tests (test_max_handler_base.py). The implementation follows established patterns from Deadline 10's render element system, ensuring compatibility and reliability.

### Was this change documented?

[...]

PR #167, opened 2025-10-27
feat(render elements): VRay split frame buffer and VFB fixes to support Render Elements

leongdl:
- Please read - https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/153 for a full code level walk through of this PR.
- This review is a chunked subset of changes specifically for the submitter UI. This is PR 4 of 4 for the render elements feature.

PR1 - Share utils: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/160
PR2 - Submission UI: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/161
PR3 - Adapter https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/162
PR4 - VRay render layer and VFB setup. This PR:  https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/167
- Original prototype (https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/152)

The 3ds Max adaptor needs to support Render Elements, and specifically for the VRay renderer.
- To use VRay with Render Elements, we need to turn off the VRay frame buffer and fallback to the 3dsmax frame buffer.
- Also, for general render elements, we did not set the output folder, enabled render elements only rendered to user configured locations. 

Implemented comprehensive V-Ray render element support with renderer detection and split buffer functionality:

**V-Ray Split Buffer Implementation:**
- Added renderer-level split buffer configuration (`output_splitgbuffer`, `output_splitRGB`, `output_splitAlpha`)
- Implemented proper split buffer filename management with base filename setting
- Added split buffer validation and output file prediction

**Enhanced V-Ray VFB Control:**
- Added global V-Ray VFB control (`rt.renderers.current.output_on = False`)
- Implemented automatic render element enabling/disabling based on ignore lists
- Added per-element VFB control with proper `vrayVFB` property management

**Renderer Agnostic Architecture:**
- Added V-Ray detection using regex pattern (`^V_Ray.*$`)
- Implemented separate workflows for V-Ray vs standard renderers
- Added standard render element filename configuration for non-V-Ray renderers

- RenderElements renders with VRay works properly even in headless mode!
- RenderElements (ART, Scanline) render elements outputs to the general job output folder.

- Created Test Bundles, rendered both using ART and VRay with render elements. Render elements are outputed as part of the job.

- NA
- Once this is merged, I am going to write a feature user guide doc.

**No, this is not a breaking change.**

- All existing parameters and APIs remain unchanged
- Backward compatible with existing job templates and UI configurations
- Graceful handling of missing parameters with appropriate defaults
- Non-V-Ray renderers get enhanced standard render element support without disruption
- Existing V-Ray workflows are enhanced but not modified in breaking ways

----

### DIFF
--- a/src/deadline/max_adaptor/MaxClient/render_element_manager.py
+++ b/src/deadline/max_adaptor/MaxClient/render_element_manager.py
@@ -5,6 +5,7 @@ Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 """
 
 import logging
+import os
 from dataclasses import dataclass
 from typing import Any, Dict, List, Optional
 
@@ -17,6 +18,8 @@ from deadline.max_shared.utilities.max_utils import (
     RenderElementInfo,
     RenderElementState,
     VRayRenderElementSettings,
+    _configure_render_element_outputs_filename,
+    _is_renderer_vray,
     configure_render_element_paths,
     configure_vray_render_elements,
     get_render_elements,
@@ -67,15 +70,34 @@ class RenderElementManager:
     - Original state storage and restoration
     """
 
-    def __init__(self):
+    def __init__(
+        self,
+        output_file_path: Optional[str] = None,
+        output_file_name: Optional[str] = None,
+        output_file_format: Optional[str] = None,
+    ) -> None:
         """
         Initialize the render element manager.
+
+        If any of the output_file_* parameters are None, they will default to the current
+        settings of the render element in the scene. Note that paths may not be path mapped
+        in this case. By default, the submitter always populates these values based on the scene.
+
+        Args:
+            output_file_path: Output directory path for split buffer
+            output_file_name: Output filename for split buffer
+            output_file_format: Output file format/extension
         """
         self.logger = logging.getLogger(__name__)
         self.re_manager: Optional[Any] = None
         self.original_state: Optional[RenderElementState] = None
-        self.cached_settings = {}
+        self.cached_settings: Dict[str, Any] = {}
         self.is_configured = False
+        self.output_file_path: Optional[str] = output_file_path
+        self.output_file_name: Optional[str] = output_file_name
+        self.output_file_format: Optional[str] = output_file_format
+        self.is_vray: bool = False
+        self.current_renderer: str = ""
 
     def _print_render_element_debug_info(self, render_elements: List[RenderElementInfo]) -> None:
         """
@@ -160,6 +182,11 @@ class RenderElementManager:
             self.original_state = store_original_render_element_state(render_elements)
             self.logger.debug(f"Stored original state for {len(render_elements)} render elements")
 
+            # Parse ignore list once for all operations
+            ignore_list = self._get_ignore_list(data)
+            if ignore_list:
+                self.logger.info(f"Ignoring render elements by name: {ignore_list}")
+
             # Configure render elements active state
             elements_enabled = self._configure_render_elements_active(data)
 
@@ -174,13 +201,35 @@ class RenderElementManager:
                 return RenderElementResult(success=True, message="Render elements disabled")
 
             # Handle ignore settings
-            self._handle_ignore_settings(data, render_elements)
+            self._handle_ignore_settings(data, render_elements, ignore_list)
 
-            # Update paths and filenames if requested
-            self._update_paths_and_filenames(data, render_elements)
+            # Detect renderer type and store as attributes
+            self.is_vray, self.current_renderer = _is_renderer_vray()
+            self.logger.info(
+                f"Detected renderer: '{self.current_renderer}' (V-Ray: {self.is_vray})"
+            )
 
-            # Handle V-Ray specific settings
-            self._configure_vray_settings(data, render_elements)
+            # Configure render element outputs based on renderer type
+            if self.is_vray:
+                # Handle V-Ray specific settings
+                self._configure_vray_settings(data, render_elements, ignore_list)
+            else:
+                # Configure standard render element outputs for non-VRay renderers
+                self.logger.info(
+                    f"Configuring standard render element outputs for '{self.current_renderer}'"
+                )
+
+                # Configure standard render element outputs
+                warnings = _configure_render_element_outputs_filename(
+                    render_elements,
+                    self.output_file_path,
+                    self.output_file_name,
+                    self.output_file_format,
+                    ignore_list,
+                )
+                if warnings:
+                    for warning in warnings:
+                        self.logger.warning(f"Standard render element configuration: {warning}")
 
             # Validate final configuration
             settings = self._convert_data_to_settings(data)
@@ -249,7 +298,7 @@ class RenderElementManager:
         return elements_enabled
 
     def _handle_ignore_settings(
-        self, data: Dict[str, Any], render_elements: List[RenderElementInfo]
+        self, data: Dict[str, Any], render_elements: List[RenderElementInfo], ignore_list: List[str]
     ) -> None:
         """
         Handle render element ignore settings.
@@ -259,15 +308,11 @@ class RenderElementManager:
             render_elements: List of render elements from scene
         """
         # Handle ignore by name list
-        ignore_names_str = self._get_param_value(data, ["ignore_render_elements_by_name"], "")
-        if ignore_names_str:
-            ignore_names = [name.strip() for name in ignore_names_str.split(",") if name.strip()]
-            self.logger.info(f"Ignoring render elements by name: {ignore_names}")
-
+        if ignore_list:
             disabled_count = 0
             for element in render_elements:
                 element_name = element.name
-                if element_name in ignore_names:
+                if element_name in ignore_list:
                     try:
                         # Disable the render element using the underlying pymxs object
                         element_obj = element.element_object
@@ -326,7 +371,7 @@ class RenderElementManager:
             raise
 
     def _configure_vray_settings(
-        self, data: Dict[str, Any], render_elements: List[RenderElementInfo]
+        self, data: Dict[str, Any], render_elements: List[RenderElementInfo], ignore_list: List[str]
     ) -> None:
         """
         Configure V-Ray specific render element settings.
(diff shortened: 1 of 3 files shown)


src/deadline/max_adaptor/MaxClient/render_element_manager.py
```diff
@@ -0,0 +1,605 @@
...
+            if self.is_vray:
+                # Handle V-Ray specific settings
+                self._configure_vray_settings(data, render_elements, ignore_list)
+            else:
+                # Configure standard render element outputs for non-VRay renderers
+                self.logger.info(
+                    f"Configuring standard render element outputs for '{self.current_renderer}'"
+                )
+
+                # Configure standard render element outputs
```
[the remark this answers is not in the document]
leongdl: old comment.

src/deadline/max_shared/utilities/max_utils.py
```diff
@@ -396,7 +386,11 @@ def configure_render_element_paths(
 
 
 def configure_vray_render_elements(
-    render_elements: list[RenderElementInfo], settings: VRayRenderElementSettings
+    render_elements: list[RenderElementInfo],
```
leongdl, line 389: Consolidated all VRay specific settings to 1 function to reduce the if-else cyclomatic complexity everywhere.

src/deadline/max_shared/utilities/max_utils.py
```diff
@@ -416,17 +416,135 @@ def configure_vray_render_elements(
     if not render_elements:
         return warnings
 
+    # Check if current renderer matches V-Ray pattern (^V_Ray.*$)
+    is_vray, current_renderer = _is_renderer_vray()
+
+    if not is_vray:
```
leongdl, line 425: This is purely for protection since the RenderElementsManager already does this check. In case the code path gets refactored.

src/deadline/max_shared/utilities/max_utils.py
```diff
@@ -416,17 +416,135 @@ def configure_vray_render_elements(
...
+    )
+
     vfb_control: bool = settings.vray_render_elements_vfb_control
+    split_buffer: bool = settings.vray_split_buffer_support
 
     try:
+        # Configure global V-Ray VFB control if enabled
+        if vfb_control:
+            try:
+                rt.renderers.current.output_on = False
```
leongdl, line 442: This is key to make VRay render buffer output rendering work.

src/deadline/max_shared/utilities/max_utils.py
```diff
@@ -416,17 +416,135 @@ def configure_vray_render_elements(
...
+
+                rt.renderers.current.output_splitRGB = True
+                rt.renderers.current.output_splitAlpha = True
+
+                _logger.info("V-Ray split buffer configured")
+            except Exception as e:
+                warnings.append(f"Failed to configure V-Ray split buffer: {e}")
+
+        # Configure split buffer filenames - set same base filename for all render elements
+        if split_buffer:
```
leongdl, line 488: Yes I'm aware they have the same `if split_buffer`, more so for refactoring in the future so it is isolated.

src/deadline/max_shared/utilities/max_utils.py
```diff
@@ -416,17 +416,135 @@ def configure_vray_render_elements(
...
+                warnings.append(f"Failed to configure global V-Ray VFB control: {e}")
+
+        # Configure split buffer if enabled
+        if split_buffer:
+            try:
+                rt.renderers.current.output_splitgbuffer = True
+
+                # Set the base filename for split files (critical for split buffer to work)
+                if output_path and output_name:
+                    base_filename = os.path.join(output_path, output_name)
```
jericht: nit: isn't this the full file path? base_filename kinda makes it sounds like just the file name part (e.g. what is returned by [os.path.basename](https://docs.python.org/3/library/os.path.html#os.path.basename)). I see V-Ray will automatically append more to this for the final file names, maybe it can be `base_filepath`?

src/deadline/max_adaptor/MaxClient/render_element_manager.py
```diff
@@ -351,15 +382,30 @@ def _configure_vray_settings(
...
             if vfb_control or split_buffer:
                 self.logger.info(
-                    f"Configuring V-Ray settings - VFB Control: {vfb_control}, Split Buffer: {split_buffer}"
+                    f"Configuring V-Ray settings for '{self.current_renderer}' - VFB Control: {vfb_control}, Split Buffer: {split_buffer}"
                 )
+                if split_buffer:
+                    if self.output_file_path and self.output_file_name:
+                        self.logger.info(
+                            f"Split buffer output: {self.output_file_path}/{self.output_file_name}"
+                        )
```
jericht, line 405: nit: should we do `os.path.join` here? Especially since 3ds Max usually runs on Windows 😄 
```suggestion
                        self.logger.info(
                            f"Split buffer output: {os.path.join(self.output_file_path, self.output_file_name)}"
                        )
```

jericht (dismissed):

src/deadline/max_adaptor/MaxClient/render_element_manager.py
leongdl: Good point - let me change that since the comment can be confusing for the reader.

src/deadline/max_shared/utilities/max_utils.py
leongdl: That's true, its the path + file name, and VRay appends the layer stuff on top. Let me refactor it.

leongdl: I have renamed it to be `base_filepath`

src/deadline/max_shared/utilities/max_utils.py
```diff
@@ -416,17 +419,157 @@ def configure_vray_render_elements(
...
+                    assert (
+                        output_file_format is not None
+                    )  # Should never be None due to default value
+                    extension = (
+                        output_file_format
+                        if output_file_format.startswith(".")
+                        else f".{output_file_format}"
+                    )
+                    filename_with_format = f"{base_name}{extension}"
+                    base_filepath = os.path.join(output_path, filename_with_format)
```
leongdl, line 467: Renamed from `base_filename` as per Jericho's suggestion.

jericht (dismissed):

src/deadline/max_adaptor/MaxClient/render_element_manager.py
```diff
@@ -5,6 +5,7 @@
 """
```
jairaws, line 5: There is some conflicting info in the description: "This is PR 2 of 3 for the render elements feature." However, below it says this is PR 4.

src/deadline/max_adaptor/MaxClient/render_element_manager.py
```diff
@@ -67,15 +70,30 @@ class RenderElementManager:
     - Original state storage and restoration
     """
 
-    def __init__(self):
+    def __init__(
```
jairaws, line 73: Can we add unit tests for all these changes?

src/deadline/max_adaptor/MaxClient/render_element_manager.py
```diff
@@ -67,15 +70,30 @@ class RenderElementManager:
...
+        self,
+        output_file_path: Optional[str] = None,
+        output_file_name: Optional[str] = None,
+        output_file_format: Optional[str] = None,
```
jairaws, line 77: What's the behaviour in case these values are not provided? It should be called out in the constructor documentation

src/deadline/max_shared/utilities/max_utils.py
```diff
@@ -396,7 +386,12 @@ def configure_render_element_paths(
 
 
 def configure_vray_render_elements(
-    render_elements: list[RenderElementInfo], settings: VRayRenderElementSettings
+    render_elements: list[RenderElementInfo],
+    settings: VRayRenderElementSettings,
+    output_path: Optional[str] = None,
+    output_name: Optional[str] = None,
+    output_file_format: Optional[str] = ".png",
```
jairaws, line 393: In this function we are defaulting the format to `.png`, but in `_configure_render_element_outputs_filename` we are defaulting it to `.exr`. Can this discrepancy create issues? Should we default to one extension only?

src/deadline/max_adaptor/MaxClient/render_element_manager.py
leongdl: Oops, my bad for copy pasting across so many PRs.

src/deadline/max_adaptor/MaxClient/render_element_manager.py
leongdl: Sure, I can add it to the docs, they are optional purely because of the job template having optional. The behavior is nothing will render and its our bug :)

src/deadline/max_adaptor/MaxClient/render_element_manager.py
leongdl: I cannot - because they all use pyxms and under the hood stuff. That's why I'm adding integ test that actually test it.

src/deadline/max_shared/utilities/max_utils.py
leongdl: A bit different in DCCs, the default exr seemed to work well with scanline. EXR with VRay didn't work when I tested it, but png does, which is why VRay is different.

Regardless, the job template always has the file type. This is just a backup that I know works well in case the user accidentally deleted the file type from the template.

src/deadline/max_adaptor/MaxClient/render_element_manager.py
leongdl: Chatted with Jair, i'll add unit tests, but maybe as a separate PR before merging this one.

Added here along with integration tests:
https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/169

(View only the test commit since its on a branch)

src/deadline/max_adaptor/MaxClient/render_element_manager.py
leongdl: Added more docs on how it will work.

jericht (approved):

AlexTranAmz (approved):

PR #182, opened 2026-01-07
feat: add VRay Lightmix integration test case

leongdl:
# Companion change to: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/183

- Customers have asked to support 3dsMax with VRay and rendering the lightmix render element with every light as a split file. This allows users to do post processing and composition by adjusting each light layer.

- No new feature toggles, we were only missing the output path setting specifically in the VRay settings struct. Before, we were setting the generic 3dsmax setting only.
- Added a new "presets" drop down with the correct setting for VRay FBO (lightmix case) and 3dsmax FBO (split elements, maxbatch case)

- Testing! Adds an integ testNo 

- I ran the bundle integ tests locally 

- NA, tests only.

- NA.

----


test/integ/test_3dsmax_render_elements.py
```diff
@@ -28,9 +28,10 @@ class TestRenderElementsAdaptors:
     @pytest.mark.parametrize(
         "bundle_dir,scene_max,max_diff_files",
         [
-            ("vray_re_test", "fog.max", 3),
+            ("vray_re_test", "fog.max", 6),
```
leongdl, line 31: Vray was really flaky because some files use random sampling. There's 60 output layers, so I relaxed it a bit.

seant-aws (approved):
Random sampling flakiness makes sense, how can one better determine flaky vs bad though?

godobyte (approved):

test/integ/test_3dsmax_render_elements.py
leongdl: Yes and no answer - because VRay is random sampling as long as all 60 files are created we are technically OK. But the render tolerance is a way to say "most" of the files are similar.

leongdl:
> Random sampling flakiness makes sense, how can one better determine flaky vs bad though?

This test mainly checks for the number of files. EG: each light is an output file. That won't be flaky. 

The light render will be varying depending on how many rays are shot. I kept the number of samples low to reduce the test time, so this is "OK" for the light mix integ case.

PR #183, opened 2026-01-07
feat: Fix Lightmix rendering with FBO split buffers

leongdl:
- Customers have asked to support 3dsMax with VRay and rendering the lightmix render element with every light as a split file. This allows users to do post processing and composition by adjusting each light layer.

- No new feature toggles, we were only missing the output path setting specifically in the VRay settings struct. Before, we were setting the generic 3dsmax setting only.
- Added a new "presets" drop down with the correct setting for VRay FBO (lightmix case) and 3dsmax FBO (split elements, maxbatch case)

- Enable users to use VRay with 3dsmax on deadline cloud better!
- Make it easier for users to use our submitter, especially when there are so many toggles.

Screenshot 2026-01-07 at 10 38 41 AM

- Created a Vray scene based on 3dsmax samples, added a VRay camera, Vray lights
- Exported a job bundle, and added an integ test - https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/182 
  - I split up the test because the test is a bunch of large binaries.
- The integ test verifies the different lights are rendered as unique files.

- NA, I will go update the user docs separately.

- No.

----

### DIFF
--- a/src/deadline/max_shared/utilities/max_utils.py
+++ b/src/deadline/max_shared/utilities/max_utils.py
@@ -92,6 +92,51 @@ class RenderElementState:
         self.vray_vfb_states = []
 
 
+# V-Ray RT class IDs (GPU renderer)
+_VRAY_RT_CLASS_IDS: list[str] = [
+    "#(1770671000, 1323107829)",
+    "#(1770671000L, 1323107829L)",
+]
+
+
+def _is_vray_rt() -> bool:
+    """
+    Check if current renderer is V-Ray RT (GPU).
+
+    Uses both class ID detection (reliable) and name-based detection (fallback).
+
+    :returns: True if V-Ray RT, False otherwise
+    :return_type: bool
+    """
+    renderer: Any = rt.renderers.current
+    renderer_class_id: str = str(renderer.classid)
+    renderer_name: str = str(renderer)
+
+    # Primary: Class ID detection (most reliable)
+    class_id_match: bool = any(class_id in renderer_class_id for class_id in _VRAY_RT_CLASS_IDS)
+
+    # Fallback: Name-based detection
+    name_match: bool = "V_Ray_GPU" in renderer_name
+
+    return class_id_match or name_match
+
+
+def _get_vray_rt_settings() -> Optional[Any]:
+    """
+    Get the V-Ray RT settings object if current renderer is V-Ray RT.
+
+    V-Ray RT (GPU) uses nested V_Ray_settings object.
+    Standard V-Ray uses direct renderer access.
+
+    :returns: V_Ray_settings object for V-Ray RT, None for standard V-Ray
+    :return_type: Optional[Any]
+    """
+    if _is_vray_rt():
+        renderer: Any = rt.renderers.current
+        return renderer.V_Ray_settings
+    return None
+
+
 def get_render_elements() -> list[RenderElementInfo]:
     """
     Gets all render elements present in the max scene with their properties.
@@ -385,6 +430,48 @@ def configure_render_element_paths(
     return warnings
 
 
+def set_vray_output_path(
+    output_path: str,
+    output_name: str,
+    output_format: str = ".exr",
+) -> list[str]:
+    """
+    Set V-Ray split buffer output path for both standard V-Ray and V-Ray RT.
+
+    Only sets the output path (output_splitfilename).
+    Split buffer flags are controlled by configure_vray_render_elements().
+
+    :param output_path: base output directory path
+    :type output_path: str
+    :param output_name: base output filename
+    :type output_name: str
+    :param output_format: output file format/extension (default: .exr)
+    :type output_format: str
+    :returns: list of warning messages
+    :return_type: list[str]
+    """
+    warnings: list[str] = []
+    split_filepath: str = os.path.join(output_path, f"{output_name}{output_format}")
+
+    try:
+        # Always set for standard V-Ray
+        rt.renderers.current.output_splitfilename = split_filepath
+
+        # Also set for V-Ray RT if applicable
+        vray_rt_settings: Optional[Any] = _get_vray_rt_settings()
+        if vray_rt_settings is not None:
+            vray_rt_settings.output_splitfilename = split_filepath
+
+        _logger.info(f"V-Ray output path set to: {split_filepath}")
+
+    except Exception as e:
+        error_msg: str = f"Failed to set V-Ray output path: {e}"
+        _logger.error(error_msg)
+        warnings.append(error_msg)
+
+    return warnings
+
+
 def configure_vray_render_elements(
     render_elements: list[RenderElementInfo],
     settings: VRayRenderElementSettings,
@@ -442,7 +529,14 @@ def configure_vray_render_elements(
         # Configure global V-Ray VFB control if enabled
         if vfb_control:
             try:
+                # Always set for standard V-Ray
                 rt.renderers.current.output_on = False
+
+                # Also set for V-Ray RT if applicable
+                vray_rt_settings: Optional[Any] = _get_vray_rt_settings()
+                if vray_rt_settings is not None:
+                    vray_rt_settings.output_on = False
+
                 _logger.info(
                     "Disabled V-Ray VFB (output_on = False) - render elements will use 3ds Max framebuffer"
                 )
@@ -452,6 +546,7 @@ def configure_vray_render_elements(
         # Configure split buffer if enabled
         if split_buffer:
             try:
+                # Always set for standard V-Ray
                 rt.renderers.current.output_splitgbuffer = True
 
                 # Set the base filename for split files (critical for split buffer to work)
@@ -480,8 +575,37 @@ def configure_vray_render_elements(
                         f"Split buffer enabled but missing: {', '.join(missing_params)} - split files may not save correctly"
                     )
 
+                # Always set for standard V-Ray
                 rt.renderers.current.output_splitRGB = True
                 rt.renderers.current.output_splitAlpha = True
+
+                # Set output_splitbitmap to enable render elements to inherit the output path
+                # This is required for V-Ray render elements (like LightMix) to save correctly
+                # The bitmap needs to be created with the split filename
+                try:
+                    if output_path and output_name:
+                        # Create a bitmap for the split output
+                        split_bitmap = rt.bitmap(1, 1, filename=base_filepath)
+                        rt.renderers.current.output_splitbitmap = split_bitmap
+                        _logger.info(f"V-Ray output_splitbitmap set to: {base_filepath}")
+                except Exception as bitmap_e:
+                    _logger.warning(f"Could not set output_splitbitmap: {bitmap_e}")
+
+                # Also set for V-Ray RT if applicable
+                vray_rt_split_settings: Optional[Any] = _get_vray_rt_settings()
+                if vray_rt_split_settings is not None:
+                    vray_rt_split_settings.output_splitgbuffer = True
+                    vray_rt_split_settings.output_splitRGB = True
+                    vray_rt_split_settings.output_splitAlpha = True
+                    # Set output_splitbitmap for V-Ray RT as well
+                    try:
+                        if output_path and output_name:
+                            split_bitmap_rt = rt.bitmap(1, 1, filename=base_filepath)
+                            vray_rt_split_settings.output_splitbitmap = split_bitmap_rt
+                    except Exception as bitmap_rt_e:
+                        _logger.warning(
+                            f"Could not set output_splitbitmap for V-Ray RT: {bitmap_rt_e}"
+                        )
 
                 _logger.info("V-Ray split buffer configured")
             except Exception as e:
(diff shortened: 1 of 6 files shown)


src/deadline/max_shared/utilities/max_utils.py
```diff
@@ -92,6 +92,51 @@ def __init__(self):
         self.vray_vfb_states = []
 
 
+# V-Ray RT class IDs (GPU renderer)
+_VRAY_RT_CLASS_IDS: list[str] = [
```
leongdl, line 96: This came from The deadline 10 porting, and its an interesting way 3dsmax encodes versions.

src/deadline/max_shared/utilities/max_utils.py
```diff
@@ -92,6 +92,51 @@ def __init__(self):
...
+
+    V-Ray RT (GPU) uses nested V_Ray_settings object.
+    Standard V-Ray uses direct renderer access.
+
+    :returns: V_Ray_settings object for V-Ray RT, None for standard V-Ray
+    :return_type: Optional[Any]
+    """
+    if _is_vray_rt():
+        renderer: Any = rt.renderers.current
+        return renderer.V_Ray_settings
```
leongdl, line 134: This is the special knowledge I had to extract from D10, VRay has a special `V_Ray_settings` object under the hood we have to setup.

src/deadline/max_shared/utilities/max_utils.py
jairaws: Does this have a human readable meaning? If so, it would be good to add a comment to each item explaining.

src/deadline/max_shared/utilities/max_utils.py
```diff
@@ -385,6 +430,48 @@ def configure_render_element_paths(
...
+    output_format: str = ".exr",
+) -> list[str]:
+    """
+    Set V-Ray split buffer output path for both standard V-Ray and V-Ray RT.
+
+    Only sets the output path (output_splitfilename).
+    Split buffer flags are controlled by configure_vray_render_elements().
+
+    :param output_path: base output directory path
+    :type output_path: str
```
jairaws: nit: Feels redundant, given that the type is defined in the interface.

src/deadline/max_shared/utilities/max_utils.py
```diff
@@ -385,6 +430,48 @@ def configure_render_element_paths(
...
+        rt.renderers.current.output_splitfilename = split_filepath
+
+        # Also set for V-Ray RT if applicable
+        vray_rt_settings: Optional[Any] = _get_vray_rt_settings()
+        if vray_rt_settings is not None:
+            vray_rt_settings.output_splitfilename = split_filepath
+
+        _logger.info(f"V-Ray output path set to: {split_filepath}")
+
+    except Exception as e:
```
jairaws, line 446: Should we throw an exception? Why are we failing silently? Will this error we caught by the logs parser?

test/unit/deadline_shared_for_3ds_max/test_shared_max_utils.py
```diff
@@ -542,3 +542,124 @@ def test_configure_render_element_outputs_filename_with_special_characters(mock_
 
     for i, expected_filename in enumerate(expected_filenames):
         mock_re_manager.SetRenderElementFilename.assert_any_call(i, expected_filename)
+
```
jairaws, line 545: Should we have tests for reading the values as well?

src/deadline/max_shared/utilities/max_utils.py
leongdl: No human readable meaning, these seems to be constants used by VRay as an internal identifier. There is not a specific thing to add as a comment other than the current comment.

src/deadline/max_shared/utilities/max_utils.py
leongdl: AI generated code docs, sure, I can remove it.

src/deadline/max_shared/utilities/max_utils.py
```diff
@@ -92,6 +92,51 @@ def __init__(self):
...
+    """
+    renderer: Any = rt.renderers.current
+    renderer_class_id: str = str(renderer.classid)
+    renderer_name: str = str(renderer)
+
+    # Primary: Class ID detection (most reliable)
+    class_id_match: bool = any(class_id in renderer_class_id for class_id in _VRAY_RT_CLASS_IDS)
+
+    # Fallback: Name-based detection
+    name_match: bool = "V_Ray_GPU" in renderer_name
```
godobyte, line 118: What's the relationship between class_id_match and name_match, would the fallback lead to false positive?

src/deadline/max_shared/utilities/max_utils.py
```diff
@@ -385,6 +430,48 @@ def configure_render_element_paths(
...
+    :type output_format: str
+    :returns: list of warning messages
+    :return_type: list[str]
+    """
+    warnings: list[str] = []
+    split_filepath: str = os.path.join(output_path, f"{output_name}{output_format}")
+
+    try:
+        # Always set for standard V-Ray
+        rt.renderers.current.output_splitfilename = split_filepath
```
godobyte, line 437: Q - What does it do for standard vray? Not having it seems fine previously?

src/deadline/max_shared/utilities/max_utils.py
godobyte: +1, would it lead to unexpected renders? Should we have logs to show whether it fails at setting standard vray or vray RT?

leongdl: This is more of a developer harness that should never be hit, unless VRay fundamentally changes the API. Thats why I did not throw an error.

If the properties for `output_splitfilename` failed to be set given a `_get_vray_rt_settings` then that means it is an API change from Chaos. 

That's why I left it as a warning that will be printed during the render.

leongdl: I did similar handling below too.

test/unit/deadline_shared_for_3ds_max/test_shared_max_utils.py
leongdl: No - because the render object is pyxms and is mocked. To read it from the render object means we have to set it. This is more of a unit test to test the code is executing as per the design.

Only our integration tests indicate this is all set correctly and rendered.

src/deadline/max_shared/utilities/max_utils.py
leongdl: The integration tests are the main level of defence here in case the underlying APIs change.

src/deadline/max_shared/utilities/max_utils.py
```diff
@@ -480,9 +575,38 @@ def configure_vray_render_elements(
...
+                # This is required for V-Ray render elements (like LightMix) to save correctly
+                # The bitmap needs to be created with the split filename
+                try:
+                    if output_path and output_name:
+                        # Create a bitmap for the split output
+                        split_bitmap = rt.bitmap(1, 1, filename=base_filepath)
+                        rt.renderers.current.output_splitbitmap = split_bitmap
+                        _logger.info(f"V-Ray output_splitbitmap set to: {base_filepath}")
+                except Exception as bitmap_e:
+                    _logger.warning(f"Could not set output_splitbitmap: {bitmap_e}")
```
godobyte, line 562: This function looks pretty long, there are more than one `if split_buffer` conditions from both above and below. This nested condition seems to appear above too

```
if split_buffer:
  try:
    if output_path and output_name:
```

Curious why is this function organize in this way.

src/deadline/max_shared/utilities/max_utils.py
jairaws: What's the UX if this fails? How does this impact customers?

src/deadline/max_submitter/ui/render_elements_widget.py
```diff
@@ -254,6 +273,51 @@ def _refresh_ui_state(self):
...
+            preset_values.get("render_elements_include_name_in_filename", True)
+        )
+        self.include_type_in_filename_checkbox.setChecked(
+            preset_values.get("render_elements_include_type_in_filename", False)
+        )
+        self.vray_vfb_control_checkbox.setChecked(
+            preset_values.get("vray_render_elements_vfb_control", True)
+        )
+        self.vray_split_buffer_checkbox.setChecked(
+            preset_values.get("vray_split_buffer_support", True)
```
godobyte, line 314: Curious is there any reasoning how the preset values are chosen? Do they worth being in comments?

src/deadline/max_shared/utilities/max_utils.py
leongdl: The class ID match came from Deadline 10, so I ported it as this constant seems to be more stable. The Name match is what we used in Deadline cloud, but I had observed some odditity depending on VRay version. Sometimes it had more `_`, and our regex'es failed.

So TLDR; I would like to migrate to the class ID in the long term just like D10. Was more reliable based on historical knowledge.

src/deadline/max_shared/utilities/max_utils.py
leongdl: Before this change, we did not set this at all, so the VRay code paths never had any split buffer setting. 

VRay also has a CPU mode, so this change makes sure any feature with or without GPU can perform split file name to the best of its capability

src/deadline/max_shared/utilities/max_utils.py
leongdl: I believe they get a black render, or a render without the split buffer. Which should be caught by our integ tests if / when we automate it.

src/deadline/max_submitter/ui/render_elements_widget.py
leongdl: There's a comment in the `data_const.py` for each of the settings groups

src/deadline/max_shared/utilities/max_utils.py
leongdl: Good question! I split the logic specifically by settings type. 

The first block from 547 to 612 is for the split buffer, the second block from 615 to 651 is for render element paths. Pyxms is tricky here to debug and we cannot lump them into 1 exception code path for debugging nicely. I wanted each setting to have an independent error out condition that is caught.

I'll take a task to refactor this function after getting this feature out. It is a bit complicated to refactor so i'll spin it off a specific refactor PR that is easier to review than mixing lightmix and this whole render element function.

jairaws (approved):

test/unit/deadline_shared_for_3ds_max/test_shared_max_utils.py
```diff
@@ -591,13 +590,38 @@ def test_set_vray_output_path_vray_rt(mock_rt: MagicMock) -> None:
...
     assert mock_vray_settings.output_splitfilename == expected_path
-    assert warnings == []
+
+
+@patch("deadline.max_shared.utilities.max_utils.rt")
+def test_set_vray_output_path_raises_on_failure(mock_rt: MagicMock) -> None:
+    """Test set_vray_output_path raises RuntimeError when setting path fails."""
+    import pytest
+
+    from deadline.max_shared.utilities.max_utils import set_vray_output_path
```
godobyte, line 606: nit

- pytest is imported from the top of the file
- other functions from `deadline.max_shared.utilities.max_utils` are imported at the top, this line is inconsistent with them unless it has specific reasons not to to group together.

src/deadline/max_shared/utilities/max_utils.py
godobyte: Sounds good! Splitting this into a dedicated refactoring PR will make the changes easier to review. Thanks for the detailed explanation of the current structure.

godobyte (approved):

test/unit/deadline_shared_for_3ds_max/test_shared_max_utils.py
```diff
@@ -542,3 +542,148 @@ def test_configure_render_element_outputs_filename_with_special_characters(mock_
...
+        set_vray_output_path(output_path, output_name, output_format)
+
+
+@patch("deadline.max_shared.utilities.max_utils.rt")
+def test_configure_vray_render_elements_sets_rt_settings(mock_rt: MagicMock) -> None:
+    """Test configure_vray_render_elements sets both standard and RT settings."""
+    from deadline.max_shared.utilities.max_utils import (
+        configure_vray_render_elements,
+        VRayRenderElementSettings,
+    )
```
godobyte, line 633: Similar nit about import.

src/deadline/max_shared/utilities/max_utils.py
godobyte: Got it, it's an optional configuration, we start to need it for Lightmix.

test/unit/deadline_shared_for_3ds_max/test_shared_max_utils.py
leongdl: Odd - how come hatch run lint didn't see this? I'll make some changes in the follow up PR.

test/unit/deadline_shared_for_3ds_max/test_shared_max_utils.py
leongdl: Will do in a clean up PR.

PR #188, opened 2026-01-15
feat: Implement vrmesh handling

leongdl:
### What was the problem/requirement? (What/Why)

V-Ray scenes using VRayProxy objects (.vrmesh files) require path mapping when rendering on Deadline Cloud workers. Without this, proxy file paths pointing to the original artist workstation locations would fail to resolve on render nodes.

### What was the solution? (How)

- Added a `map_path` callable attribute to `DefaultMaxHandler` that gets injected by `MaxClient` after renderer initialization
- Implemented `_apply_path_mapping()` hook in `DefaultMaxHandler` (no-op base) that's called after scene load
- Overrode `_apply_path_mapping()` in `VrayHandler` to iterate all VRayProxy objects and remap their `fileName` property using the injected path mapping function

### What is the impact of this change?

V-Ray scenes with VRayProxy objects will now correctly resolve .vrmesh file paths on Deadline Cloud workers, enabling successful renders of scenes that use proxy geometry.

### How was this change tested?
- A separate PR is being added for integration testing. I split it up because the integ test is all image files.

- Added comprehensive unit tests in `test_vray_path_mapping.py` covering:
  - No VRayProxy class available
  - No proxies in scene
  - Single and multiple proxy path mapping
  - Path unchanged scenarios
  - Mapping failure handling
  - Missing fileName attribute handling
  - Base handler behavior verification

### Was this change documented?

No documentation changes required - this is internal implementation.

### Did you modify schema files?
- [ ] Yes
- [X] No

If yes, Did you bump the `integration_data_interface_version` in `src/deadline/max_adaptor/MaxAdaptor/adaptor.py` and update the test constants?
See the "Adaptor Schema Files" section in DEVELOPMENT.md for details.
   - [ ] Yes
   - [X] No

### Is this a breaking change?
- No

----

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

----
Design doc:
# VRMesh Path Mapping Design Document

## Overview

This document describes the design for VRayProxy (.vrmesh) path mapping support in the Deadline Cloud for 3ds Max adaptor. VRayProxy objects reference external mesh files that need path remapping when rendering on Deadline Cloud workers.

## Problem Statement

V-Ray scenes using VRayProxy objects store absolute file paths to `.vrmesh` files. When these scenes are submitted to Deadline Cloud:

1. The original paths point to the artist's workstation (e.g., `C:/Assets/tree.vrmesh`)
[...]

PR #189, opened 2026-01-17
fix: fixed issue with Adaptor and added tests for VRayProxyObject path mapping for vrmesh files

leongdl:
- Customers have vrmesh(s) for geometry. The vrmesh file path is set as the source machine path. On a worker, the file paths change and need to be path mapped.

- Apply path mapping to each VRayProxyObject (PR) https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/188
- In this PR, we are adding an integration test.
  - The test will perform a render, check an image was rendered, and check the logs to look for the path mapping was done correctly.

- Good path mapping support for vrmesh scenes.

- Ran the integration test and it has passed. Also checked logs to make sure it is correct.

# VRMesh Path Mapping Test Results

### Test Command
```powershell
.\scripts\test-3dsmax-adapter-run.ps1 `
    -JobBundleDir "test/integ/test_scripts/vray_vrmesh_test_remap/expected_job_bundle" `
    -PathMappingFile "test/integ/test_scripts/vray_vrmesh_test_remap/path_mapping_rules.json" `
    -ShowOutput
```

### Path Mapping Rules Used
```json
{
  "version": "pathmapping-1.0",
  "path_mapping_rules": [
    {
      "source_path_format": "WINDOWS",
      "source_path": "C:/Users/RDP/deadline-cloud-for-3ds-max/test/integ/test_scripts/vray_vrmesh_test/scene",
      "destination_path": "C:/Users/RDP/deadline-cloud-for-3ds-max/test/integ/test_scripts/vray_vrmesh_test_remap/scene"
    },
    {
      "source_path_format": "WINDOWS",
      "source_path": "C:\\Users\\RDP\\deadline-cloud-for-3ds-max\\test\\integ\\test_scripts\\assets",
      "destination_path": "C:\\Users\\RDP\\deadline-cloud-for-3ds-max\\test\\integ\\test_scripts\\vray_vrmesh_test_remap\\scene"
    }
  ]
}
```

### Result: SUCCESS

Output image generated: `test/integ/test_scripts/vray_vrmesh_test_remap/tempout/State01_fog_000_Camera01.jpg`

---

### Scene Loading
```
2026/01/16 23:38:38 DBG: [18772] [16608] Interface8::LoadFromFile(C:\Users\RDP\deadline-cloud-for-3ds-max\test\integ\test_scripts\vray_vrmesh_test_remap\scene\fog.max)
2026/01/16 23:38:38 INF: [18772] [16608] Starting to load file: C:\Users\RDP\deadline-cloud-for-3ds-max\test\integ\test_scripts\vray_vrmesh_test_remap\scene\fog.max
2026/01/16 23:38:40 INF: [18772] [16608] Done loading file:  C:\Users\RDP\deadline-cloud-for-3ds-max\test\integ\test_scripts\vray_vrmesh_test_remap\scene\fog.max
```

### VRMesh Path Mapping (Key Evidence)
```
2026/01/16 23:38:40 DBG: [18772] [16608] Requesting Path Mapping for path 'C:\Users\RDP\deadline-cloud-for-3ds-max\test\integ\test_scripts\assets\Bench.vrmesh'.
2026/01/16 23:38:40 DBG: [18772] [16608] Mapped path 'C:\Users\RDP\deadline-cloud-for-3ds-max\test\integ\test_scripts\assets\Bench.vrmesh' to 'C:\Users\RDP\deadline-cloud-for-3ds-max\test\integ\test_scripts\vray_vrmesh_test_remap\scene\Bench.vrmesh'.
2026/01/16 23:38:40 DBG: [18772] [16608] Remapped VRayProxy 'VRayProxy001': C:\Users\RDP\deadline-cloud-for-3ds-max\test\integ\test_scripts\assets\Bench.vrmesh -> C:\Users\RDP\deadline-cloud-for-3ds-max\test\integ\test_scripts\vray_vrmesh_test_remap\scene\Bench.vrmesh
2026/01/16 23:38:40 DBG: [18772] [16608] Requesting Path Mapping for path 'C:\Users\RDP\deadline-cloud-for-3ds-max\test\integ\test_scripts\assets\Bench.vrmesh'.
2026/01/16 23:38:40 DBG: [18772] [16608] Mapped path 'C:\Users\RDP\deadline-cloud-for-3ds-max\test\integ\test_scripts\assets\Bench.vrmesh' to 'C:\Users\RDP\deadline-cloud-for-3ds-max\test\integ\test_scripts\vray_vrmesh_test_remap\scene\Bench.vrmesh'.
2026/01/16 23:38:40 DBG: [18772] [16608] Remapped VRayProxy 'VRayProxy002': C:\Users\RDP\deadline-cloud-for-3ds-max\test\integ\test_scripts\assets\Bench.vrmesh -> C:\Users\RDP\deadline-cloud-for-3ds-max\test\integ\test_scripts\vray_vrmesh_test_remap\scene\Bench.vrmesh
2026/01/16 23:38:40 DBG: [18772] [16608] VRMesh path mapping complete: 2 proxies remapped
```

### Render Completion
```
2026/01/16 23:38:43 DBG: [18772] [16608] [deadline.max_shared.utilities.max_utils] INFO: V-Ray output path set to: C:/Users/RDP/deadline-cloud-for-3ds-max/test/integ/test_scripts/vray_vrmesh_test_remap/tempout/State01_fog_###.jpg
2026/01/16 23:38:51 DBG: [18772] [16608] MaxClient: Finished Rendering Frame 0
```

- NA, this is a test
- 
- [ ] Yes
- [X] No

- No.
- 
----


src/deadline/max_submitter/default_max_job_template.yaml
```diff
@@ -98,7 +98,7 @@ steps:
             filename: start.bat
             type: TEXT
             data: |
-              3dsmax-openjd daemon start --connection-file {{Session.WorkingDirectory}}/connection.json --init-data file://{{Env.File.initData}}
+              3dsmax-openjd daemon start --connection-file {{Session.WorkingDirectory}}/connection.json --init-data file://{{Env.File.initData}} --path-mapping-rules file://{{Session.PathMappingRulesFile}}
```
leongdl, line 101: This was a bug! There was no way 3dsmax could path map in the adaptor.

leongdl: Example from nuke, https://github.com/aws-deadline/deadline-cloud-for-nuke/blob/1d4fef7f6226a44383f913f0436bc7f742fdc51d/src/deadline/nuke_submitter/default_nuke_job_template.yaml#L103

We forgot this in 3dsmax

test/integ/test_pathmapping.py
```diff
@@ -0,0 +1,377 @@
...
+        return None
+
+    # Look for any 3ds Max version directory (2024, 2025, 2026, etc.)
+    version_dirs = []
+    for version_dir in max_base_path.iterdir():
+        if version_dir.is_dir() and "64bit" in version_dir.name:
+            try:
+                year = int(version_dir.name.split()[0])
+                if 2024 <= year <= 2030:
+                    version_dirs.append((year, version_dir))
```
jericht, line 89: Does the expected directory name looks something like this?

```
%APP_DATA%/Autodesk/3dsMax/2024 64bit/...
```

jericht, line 89: Also might be helpful to put the min/max versions as constants somewhere at the top just so its clear this script only supports those versions

test/integ/test_pathmapping.py
```diff
@@ -0,0 +1,377 @@
...
+        # The destination path typically contains the full path like:
+        # C:/Users/RDP/deadline-cloud-for-3ds-max/test/integ/test_scripts/...
+        dest_path = rule.get("destination_path", "")
+
+        # Find the relative path starting from test/integ/test_scripts
+        relative_marker = "test/integ/test_scripts"
+        if relative_marker in dest_path.replace("\\", "/"):
+            # Extract everything from test/integ/test_scripts onwards
+            idx = dest_path.replace("\\", "/").find(relative_marker)
+            relative_path = dest_path.replace("\\", "/")[idx:]
```
jericht: nit: We can also use `str.split()` here with `relative_marker` as the split token to make it a one-liner

test/integ/test_pathmapping.py
```diff
@@ -0,0 +1,377 @@
...
+    Args:
+        log_path: Path to the 3ds Max log file.
+        expected_source_pattern: Pattern to match in the source path (e.g., "assets\\Bench.vrmesh").
+        expected_dest_pattern: Pattern to match in the destination path (e.g., "vray_vrmesh_remap_test\\scene\\Bench.vrmesh").
+
+    Returns:
+        Tuple of (success, list of matching log lines).
+    """
+    if not log_path.exists():
+        return False, [f"Log file not found: {log_path}"]
```
jericht: nit: We could also raise an exception here and have the function only return a list of matching lines rather than a tuple.

Looking at how this is used further down, I think this probably makes sense, since the test will fail when this function returns False anyway, exception will still cause it to fail

test/integ/test_pathmapping.py
```diff
@@ -0,0 +1,377 @@
...
+            print(f"Deleting existing log file for clean test run: {max_log_path}")
+            max_log_path.unlink()
+
+        test_file_location = script_location / bundle_dir
+        scene_location = test_file_location / TEST_SCENE_FOLDER / scene_max
+        output_path = tmp_path / OUTPUT_FOLDER
+        path_mapping_file = test_file_location / "path_mapping_rules.json"
+
+        # Get the checkout root (workspace root)
+        checkout_root = Path(__file__).parent.parent.parent
```
jericht: This is fragile. If this file gets moved around, this will need to be updated. Still not perfect but maybe we can just search up the file path for the first instance of the repository name? `deadline-cloud-for-3ds-max`? I think we probably won't be making an inner folder with that name

test/integ/test_pathmapping.py
```diff
@@ -0,0 +1,377 @@
...
+            assert len(expected_files) > 0, f"No expected images found in {expected_images_path}"
+
+            for expected_file in expected_files:
+                output_file = output_path / expected_file.name
+                assert output_file.exists(), (
+                    f"Expected output file not found: {output_file}\n"
+                    f"Expected based on: {expected_file}"
+                )
+                print(f"Verified output file exists: {output_file}")
+
```
jericht, line 358: We should raise an error here if the expected images path doesn't exist right?

test/integ/test_pathmapping.py
```diff
@@ -0,0 +1,377 @@
...
+            assert remap_count >= 1, (
+                f"Expected at least 1 VRMesh proxy to be remapped, but found {remap_count}.\n"
+                f"Check the log file: {max_log_path}"
+            )
+        else:
+            print(
+                f"Warning: Log file not found at {max_log_path or 'unknown path'}, "
+                "skipping log verification"
+            )
```
jericht, line 388: Could we just fail the test in this case? Log file should definitely exist if the adaptor ran 3dsMax right?

test/integ/test_pathmapping.py
leongdl: Yes, that's the standard 3dsmax path. Obviously the user installed it to a user specified directory this is not supported.

leongdl: This is a test script so yes / no.

test/integ/test_pathmapping.py
leongdl: Sure, easy refactor!

test/integ/test_pathmapping.py
leongdl: Good idea, let me refactor this. I used relative before since we don't know what computer the test will be run on.

test/integ/test_pathmapping.py
leongdl: We do? Line 340 does an assert if the expected output path does not exist.

test/integ/test_pathmapping.py
leongdl: Yeah, lets fail it - whoever runs the test should debug since it is likely a setup issue with 3dsmax path.

leongdl: Updated - assets now.

test/integ/test_pathmapping.py
leongdl: Agreed, it is now looking for `deadline-cloud-for-3dsmax` to determine this path.

test/integ/test_pathmapping.py
jericht: Oh I got confused. The `expected_images_path` is the path in the local repo, I thought this was the output path from the render.

jericht (commented):
Looks like you just need to format the code, LGTM

jericht (dismissed):

test/integ/test_pathmapping.py
```diff
@@ -0,0 +1,388 @@
...
+        repo_name = "deadline-cloud-for-3ds-max"
+        current_path = Path(__file__).resolve()
+        checkout_root = None
+        for parent in current_path.parents:
+            if parent.name == repo_name:
+                checkout_root = parent
+                break
+        assert checkout_root is not None, (
+            f"Could not find repository root '{repo_name}' in path: {current_path}"
+        )
```
godobyte: This got caught in CI, you may want to fix with `hatch run fmt`.

src/deadline/max_adaptor/MaxClient/render_handlers/vray_handler.py
```diff
@@ -117,9 +117,14 @@ def _apply_path_mapping(self) -> None:
         TODO: Add support for VRayFur (fur/hair geometry files)
         TODO: Add support for VRayScene (V-Ray scene files .vrscene)
         """
+        self.log_to_console("VrayHandler._apply_path_mapping called")
```
godobyte, line 120: nit- Should this be logged to console, or it's for debugging and we should remove?

test/integ/test_pathmapping.py
leongdl: Yes i fix this.

src/deadline/max_submitter/default_max_job_template.yaml
sakshie95, line 101: I don't think we should mark the PR "test" since we're fixing a bug in the adaptor and submitter code. Seems against the Naming convention guidelines. We should either split the PR or call it a `fix: Fixed issue with Adaptor and added tests for VRayProxyObject path mapping for vrmesh files`

sakshie95 (changes requested):

sakshie95 (commented):
Fyi, there's a security hotspot introduced which is easily fixable I would think

godobyte (approved):

src/deadline/max_submitter/default_max_job_template.yaml
leongdl: Sounds good, updated PR and commit.

src/deadline/max_adaptor/MaxClient/render_handlers/vray_handler.py
leongdl: I want to keep these for Krio debugging, it helps Kiro to trace through the system.

leongdl:
> Fyi, there's a security hotspot introduced which is easily fixable I would think

Make sure the regex used here, which is vulnerable to polynomial runtime due to backtracking, cannot lead to denial of service.

The regex is part of the test code to scan logs, there's no real vulnerability.

sakshie95 (approved):

Issue #184, opened 2026-01-12
Bug: Output file format always ends up being `TIF` instead of what was chosen in the submitter


muratmozdemir:
### Describe the bug

It looks like no matter what we choose `JPG` or `PNG` as the output format in the submitter in 3DSMax, the OutputFileFormat ends up being `.tif` in the job parameters

<img width="598" height="537" alt="Image" src="https://github.com/user-attachments/assets/8ed21be4-de55-43d4-8460-42b947e9589c" />

### Expected Behaviour

The job should output the filetype chosen in the submission

### Current Behaviour

The OutputFileFormat is always `.tif`

### Reproduction Steps

This is how we reproduce the issue:

- Using the latest submitter and 3DSMax 2025, submit a job and choose `JPG` as the output file type
- Submit the job and wait for the job to finish
- Download the output
- The output file extension we get is `.tif` instead of `.jpg` we submitted and in Deadline Cloud Monitor, we can see that the OutputFileFormat is `.tif` instead of `.jpg` we've chosen in the job submission configuration.

### Environment

1. Windows 11
2. 3DS Max 2025
3. Updated to latest on 01/07/2025

leongdl:
Thanks for the bug, I am trying to reproduce this and fix it. 

I am also working to support `.exr` and `.vrimg` properly with Render Element support.

leongdl:
I believe Zain is working on this and will have a fix soon.

ZainAallii:
Here's the PR: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/195

leongdl:
Fix is now merged - It will be part of the next release. Tentatively week off 1/26 along with other fixes.

leongdl:
This is now pending release, if this is urgently needed, please let us know and we can share a pre-release installer.

leongdl:
The fix has been released as part of 0.1.9. It is released with the next submitter installer release on 2026-02-03.

PR #195, opened 2026-01-23
fix: output extension override from submitter

ZainAallii:
When users override the output extension in the submitter UI, the override was being ignored if `rt.rendOutputFilename` was set in the render setup. The code always extracted the extension from the render setup path.

This fix ensures that settings.output_ext (UI override) takes precedence when provided, and only falls back to the render setup extension when no override is specified.

Tested by exporting job bundle and verifying parameter.yaml contains the correct overridden output extension.

When users attempted to override the output file extension in the Deadline Cloud submitter UI (e.g., changing from `.exr` to `.jpeg`), the override was being ignored if `rt.rendOutputFilename` was set in the 3ds Max render setup. The code always extracted and used the extension from the render setup path, completely bypassing the user's explicit override value in `settings.output_ext`.

This prevented users from changing output formats at submission time without modifying their scene's render setup, which is a common workflow requirement.

Here's the Github issue link: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/issues/184

Modified `src/deadline/max_submitter/max_render_submitter.py` at lines 101 and 140 to prioritize the UI override value over the render setup extension. The fix uses a conditional expression to check if `settings.output_ext` is provided, and only falls back to extracting the extension from `rt.rendOutputFilename` when no override is specified.

**Changes:**
- Line 101: Added conditional to check UI override first (for "all state sets" mode)
- Line 140: Added conditional to check UI override first (for "single state set" mode)

Both locations now use: `settings.output_ext if settings.output_ext else os.path.splitext(output_file)[1]`

Users can now successfully override the output file extension in the submitter UI, regardless of what extension is configured in the render setup. This provides greater flexibility in submission workflows without requiring scene modifications.

**Impact:**
- Positive: Users gain the ability to change output formats at submission time
- No breaking changes: When no override is provided, behavior remains unchanged (uses render setup extension)
- Backward compatible: Existing workflows without overrides continue to work as before

1. Set up a 3ds Max scene with `.exr` output extension in the render setup
Screenshot 2026-01-23 at 8 34 01 AM
3. Opened the Deadline Cloud submitter UI
4. Overrode the output extension to `.jpeg` in the submitter
5. Exported the job bundle instead of submitting
6. Verified that `parameter_values.yaml` contained the correct overridden extension (`.jpeg`) instead of the render setup default (`.exr`) which previously did not happen. 
check the parameter.yaml files before [parameter_values_fixed.yaml](https://github.com/user-attachments/files/24823286/parameter_values_fixed.yaml) and after the change [parameter_values_with_bug.yaml](https://github.com/user-attachments/files/24823287/parameter_values_with_bug.yaml)

- [x] No

This is a bug fix that restores expected behavior. No additional documentation is needed as the UI already indicates that the output extension field is an override.

- [x] No

- [ ] Yes
- [ ] No

- [x] No

This change is backward compatible. When no override is provided, the behavior remains unchanged (uses render setup extension). When an override is provided, it now correctly takes precedence as originally intended.

----

### DIFF
--- a/src/deadline/max_submitter/max_render_submitter.py
+++ b/src/deadline/max_submitter/max_render_submitter.py
@@ -98,7 +98,10 @@ def on_create_job_bundle_callback(
                 output_dir = os.path.split(rt.rendOutputFilename)[0]
                 output_file = os.path.split(rt.rendOutputFilename)[1]
                 output_file_name = Path(output_file).stem
-                output_file_format = os.path.splitext(output_file)[1]
+                # Use UI override if provided, otherwise fall back to render setup extension
+                output_file_format = (
+                    settings.output_ext if settings.output_ext else os.path.splitext(output_file)[1]
+                )
             # If it isn't, use the UI fields data
             else:
                 output_dir = settings.output_path
@@ -137,7 +140,10 @@ def on_create_job_bundle_callback(
             output_dir = os.path.split(rt.rendOutputFilename)[0]
             output_file = os.path.split(rt.rendOutputFilename)[1]
             output_file_name = Path(output_file).stem
-            output_file_format = os.path.splitext(output_file)[1]
+            # Use UI override if provided, otherwise fall back to render setup extension
+            output_file_format = (
+                settings.output_ext if settings.output_ext else os.path.splitext(output_file)[1]
+            )
         # If it isn't, use the UI fields data
         else:
             output_dir = settings.output_path


jericht (approved):

leongdl (approved):
Thank you for the fix.

ZainAallii:
[parameter_values_fixed.yaml](https://github.com/user-attachments/files/24828490/parameter_values_fixed.yaml)
[parameter_values_with_bug.yaml](https://github.com/user-attachments/files/24828491/parameter_values_with_bug.yaml)


For some reason the uploaded files were not accessible. I am uploading them again. Thanks.

PR #201, opened 2026-02-10
fix: skip output assets from asset introspection

leongdl:

The `get_referenced_files()` function was returning all assets tracked by the 3ds Max Asset Tracking System (ATS), including render output paths (main render output, render elements, Video Post outputs). These output destinations should not be included in the referenced input files list since they are destinations, not source dependencies.

Added filtering logic using `AssetManager.getAssetId()` and `AssetManager.getAsset().getType()` to check each file's asset type and exclude `#RenderOutput` and `#VideoPost` types. The implementation fails open (includes the file if type check fails) to avoid accidentally excluding valid input assets.

Job submissions will no longer incorrectly include render output paths in the asset references, resulting in cleaner job bundles that only contain actual input dependencies.

- 7 new unit tests added covering: exclusion of RenderOutput assets, exclusion of VideoPost assets, inclusion of Bitmap/XRef/Photometric assets, graceful exception handling, and mixed asset type scenarios
- All 154 existing tests pass
- hatch build, fmt, lint all pass

### Tested with 3dsmax
- from the asset tracker I can see an output file 
Screenshot 2026-02-10 at 1 52 42 PM

- After the fix, this output file is no longer in the asset file:
```
assetReferences:
  inputs:
    directories: []
    filenames:
    - C:\Users\RDP\Users\RDP\Desktop\lightmix\MyProject_$render_element.png
    - C:\Users\RDP\deadline-cloud-for-3ds-max\test\integ\test_scripts\lightmix\scene\Grycon3.tx
    - C:\Users\RDP\deadline-cloud-for-3ds-max\test\integ\test_scripts\lightmix\scene\scene.max
  outputs:
    directories:
    - C:/Users/RDP/deadline-cloud-for-3ds-max/test/integ/test_scripts/lightmix/scene/
  referencedPaths: []
```

Yes, design document added at `docs/design/filter-output-assets.md`

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

No

----

### DIFF
--- a/docs/design/filter-output-assets.md
+++ b/docs/design/filter-output-assets.md
@@ -1,0 +1,186 @@
+# Filter Output Assets from get_referenced_files()
+
+## Problem
+
+The `get_referenced_files()` function currently uses `ATSOps.GetFiles()` which returns all tracked assets, including render outputs (main render output path, Render Elements like AO/Z-depth, and Bake-to-Texture paths). These output files should not be included in the referenced files list since they are destinations, not source dependencies.
+
+## Current Implementation
+
+```python
+def get_referenced_files() -> list:
+    rt.ATSOps.Refresh()
+    maps = rt.ATSOps.GetFiles(pymxs.byref(None))
+    maps_in_scene = [str(x.replace("\\", "/")) for x in list(maps)[1]]
+    # ... nested file handling ...
+    return maps_in_scene
+```
+
+## Proposed Change (Minimal)
+
+Use `AssetManager.getAssetId()` and `AssetManager.getAsset()` to check each file's asset type and filter out `#RenderOutput` and `#VideoPost` types.
+
+```python
+def get_referenced_files() -> list:
+    """
+    Finds all referenced files (bitmap textures, xrefs).
+    Excludes render output files (render outputs, render elements, video post).
+
+    :returns: list with paths to all referenced files
+    :return_type: list
+    """
+    # Refresh Asset Tracking to make sure we have the latest version
+    rt.ATSOps.Refresh()
+
+    # Convert result from usedMaps MAXScript function into a python list
+    maps = rt.ATSOps.GetFiles(pymxs.byref(None))
+    # Note: ATSOps returns list[int, [file paths]]
+    all_files = list(maps)[1]
+
+    # Filter out output assets (RenderOutput, VideoPost)
+    maps_in_scene = []
+    for f in all_files:
+        try:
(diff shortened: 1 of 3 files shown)


test/unit/deadline_submitter_for_3ds_max/test_max_utils.py
```diff
@@ -0,0 +1,186 @@
...
+        asset_map, create_asset = mock_asset_manager
+
+        # Setup files: one texture, one render output
+        files = ["C:\\textures\\diffuse.jpg", "C:\\output\\render.png"]
+        mock_rt.ATSOps.GetFiles.return_value = (2, files)
+
+        asset_map["C:\\textures\\diffuse.jpg"] = create_asset("#Bitmap")
+        asset_map["C:\\output\\render.png"] = create_asset("#RenderOutput")
+
+        from deadline.max_submitter.utilities.max_utils import get_referenced_files
```
kenyrish-amazon, line 64: Nit but each test has this import, we can just add it at the top of the file

docs/design/filter-output-assets.md
```diff
@@ -0,0 +1,186 @@
...
+    rt.ATSOps.Refresh()
+    maps = rt.ATSOps.GetFiles(pymxs.byref(None))
+    maps_in_scene = [str(x.replace("\\", "/")) for x in list(maps)[1]]
+    # ... nested file handling ...
+    return maps_in_scene
+```
+
+## Proposed Change (Minimal)
+
+Use `AssetManager.getAssetId()` and `AssetManager.getAsset()` to check each file's asset type and filter out `#RenderOutput` and `#VideoPost` types.
```
kenyrish-amazon, line 20: Is there a list of different Asset types somewhere? Are these the only two types we can safely remove?

kenyrish-amazon (approved):

seant-aws (approved):

docs/design/filter-output-assets.md
leongdl: Yes - see line 89 below. I google'ed it!

test/unit/deadline_submitter_for_3ds_max/test_max_utils.py
leongdl: Sorry cannot, because of import mocking:

```
The issue is that pymxs needs to be mocked before importing get_referenced_files since the module imports pymxs at load time. I need to restructure the test to use a module-level fixture that patches pymxs before the import:
```

Issue #76, opened 2025-02-05
Bug: Unable to add Dev Adaptor Wheels


yuanmich2:
### Describe the bug

When I set the dev environment variable and then run the submitter, I see a checkbox for adding the adaptor wheels, but there doesn't seem to a field for me to specify where the adaptor wheels are. When I hit submit, I get an error that says "NoneType" objects is not subscriptable

![Image](https://github.com/user-attachments/assets/e2688308-bdfd-4372-aab2-59613762f2bf)

### Expected Behaviour

A field for the adaptor wheels should have appeared and I should have been able to provide adaptor wheels.

### Current Behaviour

The submitter doesn't seem to handle adaptor wheels correctly.

### Reproduction Steps

Install the submitter and try to add adaptor wheels.

### Environment

At minimum:
1. Operating system Windows Server 2022
2. Version of 3ds Max
3. Version of this package 0.1.2

Please share other details about your environment that you think might be relevant to reproducing the bug.

AlexTranAmz:
Thanks for reporting this issue. We'll take a look to determine the best path forward.

caleb-mabry:
I am also experiencing this issue. 

Were you able to get around this? @yuanmich2

jairaws:
Hey @caleb-mabry, we need to do some refactoring to the submitter to get this feature working. There is no fix yet, but it is in our backlog.

ZainAallii:
I went past the error with some changes (which I forgot) @yuanmich2 can you share the output from Max Script Listener> Scripting> Script Listener ? Configure it like this: 

<img width="308" height="292" alt="Image" src="https://github.com/user-attachments/assets/f36ac715-384a-4245-b51f-0453d0501e8e" />

and then share the error from the listener while trying to submit. You will have to click `ok` on the UI of the error message to be able to copy the text from the listener

leongdl:
Fix has been merged - this is pending the next 3dsmax release.

leongdl:
This fix has been released as part of the 3/10 submitter release.

PR #202, opened 2026-02-11
feat: add developer option to override adaptor with local wheels

godobyte:
https://github.com/aws-deadline/deadline-cloud-for-3ds-max/issues/76

Developers working on the 3ds Max adaptor needed a way to test local adaptor changes on workers without having to publish and install new versions. This is essential for rapid iteration during development and debugging of adaptor-specific issues.

Added a developer-only "Include Adaptor Wheels" checkbox in the Scene Settings tab that:
- Only appears when `DEADLINE_ENABLE_DEVELOPER_OPTIONS=true` environment variable is set
- Allows developers to attach a `wheels/` directory containing local builds of `openjd-adaptor-runtime`, `deadline`, and `deadline-cloud-for-3ds-max` packages
- Creates a Python virtual environment on the worker and installs the custom wheels before running the adaptor
- Uses a single cross-platform Python script (`setup_adaptor_wheels.py`) that automatically handles platform differences (Windows vs Linux)

- Developers can now test local adaptor changes on workers without modifying installed packages
- Full cross-platform support for both Windows and Linux workers using a single Python script
- Cleaner, more maintainable implementation with the setup logic in a separate Python file
- No impact on production users (feature is hidden behind developer environment variable)

- Unit tests added and passing
- Manually tested on Windows worker with local adaptor wheels
- Verified checkbox only appears when `DEADLINE_ENABLE_DEVELOPER_OPTIONS=true`
- Confirmed worker creates venv, installs wheels (including dev versions with git hashes), and runs custom adaptor successfully
- Verified cross-platform Python script handles platform detection correctly

- [x] Yes

Added comprehensive documentation in `DEVELOPMENT.md` under "Running the Adaptor on a Farm" section with:
- Step-by-step instructions for building and using adaptor wheels
- Platform-specific commands for Windows (PowerShell/CMD) using environment variables
- Updated `.kiro/powers/3dsmax-dev-power/steering/build-and-test.md` with adaptor wheels workflow
- Generic path references using `$env:3DSMAX_EXECUTABLE` instead of hardcoded paths

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

- [ ] Yes
- [x] No

This is a purely additive feature that only affects developers who explicitly enable it via environment variable.

----

### DIFF
--- a/src/deadline/max_submitter/create_job_bundle.py
+++ b/src/deadline/max_submitter/create_job_bundle.py
@@ -413,7 +413,7 @@ def _create_step_definitions(
     return job_template
 
 
-def _merge_adaptor_override_environment():
+def _merge_adaptor_override_environment() -> dict[str, Any]:
     """
     Create template for the adaptor override environment.
 
@@ -450,7 +450,7 @@ def _merge_adaptor_override_environment():
     wheels_path_package_names = {path.split("-", 1)[0] for path in wheel_files}
 
     # Check for duplicate packages (multiple versions of the same package)
-    package_counts = {}
+    package_counts: dict[str, int] = {}
     for wheel_file in wheel_files:
         package_name = wheel_file.split("-", 1)[0]
         package_counts[package_name] = package_counts.get(package_name, 0) + 1
--- a/src/deadline/max_submitter/create_job_bundle.py
+++ b/src/deadline/max_submitter/create_job_bundle.py
@@ -415,46 +415,119 @@ def _create_step_definitions(
 
 def _merge_adaptor_override_environment():
     """
-    Create template for the adaptor override environment
+    Create template for the adaptor override environment.
+
+    Loads the adaptor_override_environment.yaml file, validates the wheels directory,
+    and configures the OverrideAdaptorName parameter with the default value.
+
+    :return: The override environment dictionary loaded from YAML
+    :raises RuntimeError: If the YAML file is invalid or wheels directory is missing/incorrect
     """
-    # TODO see if changes to adaptor_override_environment.yaml need to be made to work on windows worker
     with open(Path(__file__).parent / "adaptor_override_environment.yaml") as f:
         override_environment = yaml.safe_load(f)
 
-    # Read DEVELOPMENT.md for instructions to create the wheels directory.
+    if override_environment is None:
+        raise RuntimeError(
+            "Failed to load adaptor_override_environment.yaml - file is empty or invalid"
+        )
+
+    if "parameterDefinitions" not in override_environment:
+        raise RuntimeError(
+            f"adaptor_override_environment.yaml is missing 'parameterDefinitions'. Keys found: {list(override_environment.keys())}"
+        )
+
+    # Validate wheels directory exists and contains the correct packages
     wheels_path = Path(__file__).parent.parent.parent.parent / "wheels"
-    if not wheels_path.exists() and wheels_path.is_dir():
+    _logger.info(f"Validating wheels directory: {wheels_path}")
+
+    if not wheels_path.exists() or not wheels_path.is_dir():
         raise RuntimeError(
             "The Developer Option 'Include Adaptor Wheels' is enabled, "
-            "but the wheels directory does not exist:\n" + str(wheels_path)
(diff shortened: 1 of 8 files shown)


src/deadline/max_submitter/create_job_bundle.py
```diff
@@ -415,46 +415,119 @@ def _create_step_definitions(
 
 def _merge_adaptor_override_environment():
```
leongdl: nit, return type. Ask Kiro to generate all my py args and returns.

src/deadline/max_submitter/create_job_bundle.py
```diff
@@ -415,46 +415,119 @@ def _create_step_definitions(
...
     """
-    Create template for the adaptor override environment
+    Create template for the adaptor override environment.
```
leongdl, line 418: https://github.com/leongdl/deadline-cloud-for-3ds-max/blob/eaf1ba92854a2365a4f3b6a7b9cc29b180a4a99a/src/deadline/max_submitter/create_job_bundle.py#L55

Should line 56 and 57 be swapped?

We should have the override environment to create the venv, then append the job template job envs?

leongdl (changes requested):
Please double check the job enviornment merge code.

leongdl:
And CI failed on python stuff

src/deadline/max_submitter/create_job_bundle.py
godobyte: As of now, this order works fine since the 3dsmax env is a step env, which runs after the adapter override job env or any other job env. Since the default 3ds Max template has no job environments, both approaches work the same right now.

It seems logical to have adaptor override run early to set up the custom adaptor before other job-level setup that might depend on it, but I wonder whether there are other job env that needs to happen before the adapter override env. 

Additionally, Maya appends this override env to other job envs, I think we should have the same pattern across DCC adaptors.

src/deadline/max_submitter/create_job_bundle.py
```diff
@@ -413,48 +413,121 @@ def _create_step_definitions(
...
+
+    :return: The override environment dictionary loaded from YAML
+    :raises RuntimeError: If the YAML file is invalid or wheels directory is missing/incorrect
     """
-    # TODO see if changes to adaptor_override_environment.yaml need to be made to work on windows worker
     with open(Path(__file__).parent / "adaptor_override_environment.yaml") as f:
         override_environment = yaml.safe_load(f)
 
-    # Read DEVELOPMENT.md for instructions to create the wheels directory.
+    if override_environment is None:
```
leongdl, line 429: I don't think it can be none unless the file is empty since it does a with open above.  But super nit that i do not think matters.

src/deadline/max_submitter/setup_adaptor_wheels.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python3
```
leongdl, line 1: I think this file is missing a copyright header.

godobyte: Good catch, I wonder why CI didn't catch it 🤔

src/deadline/max_submitter/create_job_bundle.py
godobyte: While unlikely, I still kept it here since yaml.safe_load() returns None for empty files, whitespace-only files, or files with only comments. The check prevents potential errors if the template file gets corrupted or cleared during development.

seant-aws (approved):

src/deadline/max_submitter/setup_adaptor_wheels.py
godobyte: Ah apparently the [test_copyright_headers.py](https://github.com/aws-deadline/deadline-cloud-for-3ds-max/blob/mainline/test/test_copyright_headers.py) didn't run properly. I have the fix locally, if there are other comments I can address together, otherwise I'd like to fix this header issue separately. Fixing the test discovered more files with no headers

leongdl (approved):

Issue #116, opened 2025-07-08
Feature request: Support the Batch Render Tool


jblagden:
### Describe the problem

3dsMax has a tool to submit multiple render tasks from a single scene, typically with changes in camera or some other render settings (ex, day shot, night shot, dusk, rain, etc).

Putting all these tasks together on the farm would be helpful.

### Proposed Solution

Ideally we could choose 'submit batch render queue' at submission time, and the render settings in there would create multiple jobs or steps.

I think multiple jobs would be the simplest, but it could make sense to submit a single job for the scene file, and a step for each batch render entry.

### Example Use Cases

Ideally this would be a part of the integrated submitter. I'm noodling around with seeing how Deadline 10 did this and will update if I have anything potentially useful.

jericht:
This sounds like a good addition @jblagden! What is the tool that 3dsMax has for submitting multiple render tasks from a single scene? If it's built into 3dsMax, we could probably integrate with it directly.

jblagden:
It's [this guy](https://help.autodesk.com/view/3DSMAX/2025/ENU/?guid=GUID-21D8346E-6F77-483C-9389-203977D4A9CB) I thought I'd linked it! :D I'm sure there's a way to integrate with it directly.

leongdl:
I would like to migrate the feature request discussion to - https://github.com/aws-deadline/deadline-cloud-for-3ds-max/discussions/140#discussioncomment-14301419

This will help prioritize and drive uptake into what features should be prioritized.

leongdl:
This is now work in progress with @jericht

jericht:
Fixed in https://github.com/aws-deadline/deadline-cloud-for-3ds-max/releases/tag/0.2.0

PR #205, opened 2026-02-14
feat!: add batch rendering support

jericht:

Users needed the ability to submit batch render jobs from 3ds Max's Batch Render Manager to Deadline Cloud. See #116

This PR adds a new "Batch Render" submission mode alongside the existing "Default" (State Sets + Cameras) mode. The two modes are mutually exclusive, selected via radio buttons in the Scene Settings tab.

#### Submission modes

In **Default mode**, the submitter creates one step per State Set with per-step parameters (frames, output path/name/format, resolution). A global Camera parameter is added when a specific camera is selected. This is the same workflow as before, but parameters are now scoped per step instead of shared globally.

**Default mode GUI**

Screenshot 2026-03-19 at 9 32 30 AM

In **Batch Render mode**, the submitter reads each enabled batch view from the Batch Render Manager at submission time and creates one step per view. Each step gets its own parameters for frames, output path/name/format, resolution, camera, scene state, preset file, and pixel aspect — all extracted from the scene and baked into the job bundle. The adaptor never queries the Batch Render Manager.

**Batch Render mode GUI**

Screenshot 2026-03-19 at 9 32 48 AM

#### Submitter UI

- Default mode groups State Sets, Cameras, and output settings into one panel; Batch Render mode shows a read-only table of enabled batch views with columns for Scene State, Output Path, and Override Preset
- Batch views auto-refresh on window focus
- Table columns support interactive resizing, middle-ellipsis for long paths, and tooltips
- Parameter names are index-prefixed (e.g. `B1_MyBatchView_Frames`) to stay within OpenJD's 64-char limit
- Sanity checks validate camera existence, scene state existence, preset file existence, pixel aspect positivity, and non-empty output filenames

#### Adaptor

- New action handlers: `set_scene_state` (restores via `rt.sceneStateMgr`), `set_preset_file` (with path mapping), `set_pixel_aspect`, `set_batch_name`
- `_dump_max_log_on_error()` prints the full Max.log on render failures

#### Documentation

Added a "Known Limitations" section to `docs/user_guide/using-submitter.md` documenting the OpenJD 50-parameter ceiling:
- Default mode: max 6–8 state sets per job
- Batch Render mode: max 3–4 batch views per job

### Example Job Bundles

<details>
<summary>Default Submission Mode (2 state sets)</summary>

```yaml
parameterValues:
- name: MaxSceneFile
  value: C:/Users/RDP/Documents/3ds Max 2025/scenes/Automotive_Exterior/Automotive_Exterior.max
- name: FenderDiffColor_Frames
  value: '0'
- name: FenderDiffColor_OutputFilePath
  value: C:/Users/RDP/Documents/3ds Max 2025/renderoutput
- name: FenderDiffColor_OutputFileName
  value: <camera>_<stateset>_Car_##
- name: FenderDiffColor_OutputFileFormat
  value: .exr
- name: FenderDiffColor_ImageWidth
  value: 1920
- name: FenderDiffColor_ImageHeight
  value: 1080
- name: FenderInvisible_Frames
  value: '0'
- name: FenderInvisible_OutputFilePath
  value: C:/Users/RDP/Documents/3ds Max 2025/renderoutput
- name: FenderInvisible_OutputFileName
  value: <camera>_<stateset>_Car_##
- name: FenderInvisible_OutputFileFormat
  value: .exr
- name: FenderInvisible_ImageWidth
  value: 1920
- name: FenderInvisible_ImageHeight
  value: 1080
```
</details>

<details>
<summary>Batch Render Submission Mode (4 batch views)</summary>

```yaml
parameterValues:
- name: MaxSceneFile
  value: C:/Users/RDP/Documents/3ds Max 2025/scenes/Automotive_Exterior/Automotive_Exterior.max
- name: bus_Frames
  value: '0'
- name: bus_OutputFilePath
  value: C:/Users/RDP/Documents/3ds Max 2025/renderoutput
- name: bus_OutputFileName
  value: <camera>_<batch>_Car_##
- name: bus_OutputFileFormat
  value: .exr
- name: bus_ImageWidth
  value: 1920
- name: bus_ImageHeight
  value: 1080
- name: f12to13_r5x5_hdr_Frames
  value: 12-13
- name: f12to13_r5x5_hdr_OutputFilePath
  value: C:/Users/RDP/Documents/3ds Max 2025/renderoutput
- name: f12to13_r5x5_hdr_OutputFileName
  value: <camera>_<batch>_Car_##
- name: f12to13_r5x5_hdr_OutputFileFormat
  value: .exr
- name: f12to13_r5x5_hdr_ImageWidth
  value: 5
- name: f12to13_r5x5_hdr_ImageHeight
  value: 5
- name: f12to13_r5x5_hdr_PresetFile
  value: C:\Users\RDP\Documents\3ds Max 2025\renderpresets\jericht-the-one.rps
- name: f12to13_r5x5_hdr_PixelAspect
  value: '1.0'
- name: f5to10_r10x10_jpg_Frames
  value: 5-10
- name: f5to10_r10x10_jpg_OutputFilePath
  value: C:/Users/RDP/Documents/3ds Max 2025/renderoutput
- name: f5to10_r10x10_jpg_OutputFileName
  value: <camera>_<batch>_Car_##
- name: f5to10_r10x10_jpg_OutputFileFormat
  value: .exr
- name: f5to10_r10x10_jpg_ImageWidth
  value: 10
- name: f5to10_r10x10_jpg_ImageHeight
  value: 10
- name: f5to10_r10x10_jpg_Camera
  value: RenderCam
- name: f5to10_r10x10_jpg_PresetFile
  value: C:\Users\RDP\Documents\3ds Max 2025\renderpresets\jericht-f5to10-r10x10.rps
```
</details>

This is a breaking change. Job parameters are now generated dynamically per step instead of being defined statically in the default template. The `integration_data_interface_version` is bumped from 0.1 to 0.2. Existing job bundles will not work with this adaptor version.

- Unit tests for submitter (step generation, param definitions, sanity checks, data classes, parameter naming, batch view extraction)
- Unit tests for adaptor (new action handlers, schema validation, camera fallback, base handler)
- Updated integration test expected bundles for step-specific parameter naming
- Manual E2E testing on SMF with V-Ray 7 and Scanline renderer across both submission modes

- [x] Yes — Known Limitations section added to user guide

- [x] Yes — `integration_data_interface_version` bumped, test constants updated

Yes — job bundle format changed. Parameters are now per-step with prefixed names. New init-data keys (`scene_state`, `preset_file`, `pixel_aspect`, `batch_name`) added for batch render mode.

### Known issues

- Render elements may not produce output in Batch Render mode with V-Ray (under investigation)
- OpenJD 50-parameter limit caps batch views at 3–4 and state sets at 6–8 per job

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


seant-aws (commented):
A screenshot of the new submitter UI for batch rendering would be helpful !

src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
```diff
@@ -43,18 +44,26 @@ def __init__(self):
...
             "cleanup_render_elements": self.cleanup_render_elements,
+            # Batch render item application
+            "batch_item_name": self.apply_batch_item,
         }
         self.camera_node = None
         self.output_dir = None
         self.output_name = None
         self.output_format = None
         self.state_set_name: str = ""
+        self.batch_item_name = None  # Batch render item name (if batch rendering)
```
seant-aws: nit: do we need this comment ? the others don't have it

src/deadline/max_shared/utilities/max_utils.py
```diff
@@ -92,6 +92,150 @@ def __init__(self):
...
+    name = str(view.name) if view.name else f"BatchItem_{index}"
+    enabled = bool(view.enabled)
+
+    # Extract camera name
+    camera = None
+    if view.camera is not None and view.camera != rt.undefined:
+        camera = str(view.camera.name)
+
+    # Extract output filename
+    output_filename = str(view.outputFilename) if view.outputFilename else ""
```
seant-aws, line 144: how is this output_filename used? cuz in the default_max_handler.py, it seems like we're using the batch_item_name directly? or maybe I'm misunderstanding

src/deadline/max_shared/utilities/max_utils.py
```diff
@@ -92,6 +92,150 @@ def __init__(self):
...
+@dataclass
+class BatchItem:
+    """
+    Data class representing a batch render item from 3ds Max's Batch Render Manager.
+    """
+
+    name: str
+    enabled: bool = True
+    camera: Optional[str] = None
+    output_filename: str = ""
```
seant-aws, line 105: given the output file name, what takes precedence? the specified output filename in the submitter? or the specified output filename from the batch items menu, since we can also specify there? more of a UX question

seant-aws: as well, noted in discussion, we also want to add a fix for output filepaths / output filenames; if empty, should show user some sort of default filepath

src/deadline/max_submitter/ui/scene_settings_tab.py
```diff
@@ -30,14 +30,18 @@
     QGridLayout,
```
seant-aws, line 31: noted from in person discussion, the UI menu had some padding bug where we had to size up horizontally

test/unit/deadline_submitter_for_3ds_max/test_sanity_checks.py
```diff
@@ -225,3 +225,143 @@ def test_allowed_renderers_no_substring_conflicts(self) -> None:
...
+
+        with raises(Exception, match="No enabled batch items found"):
+            check_sanity_batch_render(batch_settings)
+
+    @patch("deadline.max_submitter.sanity_checks.get_batch_render_items")
+    def test_passes_with_enabled_items(
+        self, mock_get_batch_items, batch_settings: RenderSubmitterUISettings
+    ):
+        """Verify no exception when enabled batch items exist."""
+        from deadline.max_submitter.sanity_checks import check_sanity_batch_render
```
seant-aws: nit: is this pattern of importing per test better than importing on the top

src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
```diff
@@ -43,18 +44,26 @@ def __init__(self):
...
         self._executable_handler: MaxExecutableHandler = MaxExecutableHandler()
         # Initialize render element manager as private attribute
         self._render_element_manager: Optional["RenderElementManager"] = None
         # Path mapping function injected by MaxClient
         # Uses Callable type since map_path comes from ClientInterface
         self.map_path: Optional[Callable[[str], str]] = None
+        # Batch item applier - use lambdas to defer lookup of map_path since it's set after __init__
+        self._batch_item_applier = BatchItemApplier(
+            map_path=lambda path: self.map_path(path) if self.map_path else path,
+            set_camera_callback=lambda node: setattr(self, "camera_node", node),
```
leongdl: super nit, 3dsmax prints some odd errors if setattr fails. Is it sure we will have "camera_node" ?

src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
```diff
@@ -355,17 +392,12 @@ def _apply_path_mapping(self) -> None:
...
         """
-        # When using 3dsmaxbatch (batch mode), log to Max.log file only
-        try:
-            # When using 3dsmax (interactive mode), print to stdout
-            print(message, flush=True)
-            rt.logsystem.logEntry(message, broadcast=True)
-        except Exception:
-            # If logsystem fails, continue without breaking execution
-            pass
+        log_to_console(message)
```
leongdl, line 556: Something is odd here - log_to_console calls log_to_console ?

src/deadline/max_submitter/default_max_job_template.yaml
```diff
@@ -18,23 +18,6 @@ parameterDefinitions:
       patterns:
       - '*'
   description: The Max scene file to render.
-- name: Frames
```
leongdl, line 21: Would old job bundles fail since the new default bundle is updated?

jericht: No, the adaptor will still work since the init-data and run-data have not changed in a breaking way.

Only a new optional parameter for the batch view name was added to the init-data

src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
jericht: Yes, since I wanted to group all the batch render functions to a different class, I moved the `log_to_console` function out to a module level function since it didn't need to be a method on the class. However a derived class referenced it via `self`, so I kept this here to keep the code diff smaller.

I can update the derived class to import and call the module level function directly though

src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
jericht: Yes. `setattr` works even if the object does not have the attribute: https://docs.python.org/3/library/functions.html#setattr

> This is the counterpart of [getattr()](https://docs.python.org/3/library/functions.html#getattr). The arguments are an object, a string, and an arbitrary value. The string may name an existing attribute or a new attribute. The function assigns the value to the attribute, provided the object allows it. For example, setattr(x, 'foobar', 123) is equivalent to x.foobar = 123.
>
> name need not be a Python identifier as defined in [Names (identifiers and keywords)](https://docs.python.org/3/reference/lexical_analysis.html#identifiers) unless the object chooses to enforce that, for example in a custom [__getattribute__()](https://docs.python.org/3/reference/datamodel.html#object.__getattribute__) or via [__slots__](https://docs.python.org/3/reference/datamodel.html#object.__slots__). An attribute whose name is not an identifier will not be accessible using the dot notation, but is accessible through [getattr()](https://docs.python.org/3/library/functions.html#getattr) etc..

I guess some 3dsMax object have this custom `__get_attribute__` or `__slots__` implementation, but this class doesn't have that

test/unit/deadline_submitter_for_3ds_max/test_sanity_checks.py
jericht: No, it is only useful for optional/conditional imports. I've moved unnecessary imports inside functions to module level

src/deadline/max_submitter/ui/scene_settings_tab.py
jericht: This has since been fixed. It was an issue with the QGridLayout and some elements being in column 5 when we really only need 2

leongdl (dismissed):

src/deadline/max_shared/utilities/max_utils.py
jericht: I think the output in the submitter UI should take precedence since it's required. Will need to update the batch render UI to reflect this.

src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
leongdl: OK - thats fine. I called it out because I saw some red herrings before.

src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
leongdl: Whatever makes the code simpler to read.

jericht:
There are still a few ambiguities I want to iron out for this feature and manual tests I want to perform before merging this. I'll likely be pushing up more revisions with changes from my testing in the next couple days.

src/deadline/max_adaptor/MaxAdaptor/adaptor.py
```diff
@@ -343,6 +355,53 @@ def _populate_action_queue(self) -> None:
                 "Render element modification is disabled, skipping render element configuration"
             )
 
+    def _dump_max_log_on_error(self) -> None:
```
leongdl, line 365: Future thought - what if we added a debug flag that always prints the max.log. Its good for debugging bad renders, and /or we can copy the log back as a render result and place it somewhere meaningful.

src/deadline/max_adaptor/MaxAdaptor/adaptor.py
```diff
@@ -343,6 +355,53 @@ def _populate_action_queue(self) -> None:
...
+        """
+        import glob
+
+        local_appdata = os.environ.get("LOCALAPPDATA", "")
+        if not local_appdata:
+            _logger.warning("LOCALAPPDATA not set, cannot locate Max.log")
+            return
+
+        pattern = os.path.join(
+            local_appdata, "Autodesk", "3dsMax", "*", "ENU", "Network", "Max.log"
```
leongdl, line 385: Be careful, this works for the default location but will break for users that install it elsewhere, eg on a network disk, or the experimental conda package.

src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
```diff
@@ -120,31 +129,65 @@ def start_render(self, data: dict) -> None:
...
                 camera = self.get_camera_to_render(camera)
                 self.camera_node = rt.getNodeByName(camera)
 
-            # Since camera can be set by both init and run data, this isn't a required parameter in either schema.
+            # If no camera was set by init or run data, use the scene's active camera
             if self.camera_node is None:
-                self.log_to_console("Error: MaxClient: start_render called without a camera.")
-                raise RuntimeError("MaxClient: start_render called without a camera.")
+                cameras = list(rt.cameras)
+                if cameras:
```
leongdl, line 139: Good catch!! This was always broken

src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
```diff
@@ -120,31 +129,65 @@ def start_render(self, data: dict) -> None:
...
-            camera_name = camera or ""
+            # Camera name: prefer run data camera, fall back to init-data camera (self.camera_node)
+            if camera:
+                camera_name = camera
+            elif self.camera_node is not None:
+                camera_name = str(self.camera_node.name)
+            else:
+                camera_name = ""
 
+            import deadline.max_shared.utilities.filename_utils
```
leongdl: nit, can we move this to the top

src/deadline/max_adaptor/MaxClient/render_handlers/vray_handler.py
```diff
@@ -186,36 +186,28 @@ def start_render(self, data: dict[str, Any]) -> None:
...
-                    self.log_to_console(f"Warning: {warning}")
-            else:
-                # Standard V-Ray output path for other formats
-                set_vray_output_path(
-                    output_path=self.output_dir,
-                    output_name=resolved_name,
-                    output_format=output_format,
-                )
-
-        super().start_render(data)
```
leongdl, line 221: Did we lose this line intentionally? I don't see a start_render

src/deadline/max_adaptor/MaxClient/render_handlers/vray_handler.py
```diff
@@ -186,36 +186,28 @@ def start_render(self, data: dict[str, Any]) -> None:
...
         For other formats, uses standard V-Ray split buffer output.
         """
-        if self.output_dir and self.output_name:
-            output_format = self.output_format or ".exr"
-
-            camera_name = data.get("camera", "") or ""
-            scene_name = Path(rt.maxFileName).stem if rt.maxFileName else ""
-            state_set_name = self.state_set_name or ""
-
-            resolved_name = format_output_filename(
```
leongdl, line 196: Is this done at the parent level?

leongdl: Ok i do see it above at the default_max_handler, just cheking

src/deadline/max_adaptor/MaxClient/render_handlers/vray_handler.py
leongdl: I see, the base start_render still exists, but now calls the subclass - correct me if this is the right understanding

src/deadline/max_submitter/utilities/max_utils.py
```diff
@@ -344,7 +349,108 @@ def get_render_output_info() -> tuple[str, str, str]:
...
+    Uses the robust save_max_backup_file and restore_max_copy functions which properly
+    handle edge cases like preserving existing hold files. Guarantees restoration even
+    if an exception occurs.
+
+    Usage:
+        with hold_and_fetch_scene():
+            # Make temporary changes to scene
+            # Changes will be reverted on exit
+    """
+    backup_path = rt.execute("GetDir #temp") + "\\" + TEMP_BACKUP_FILENAME
```
leongdl, line 372: we had questions / concerns about hold and fetch. I guess this is just a carry over from the D10 architecture?

src/deadline/max_submitter/default_max_job_template.yaml
```diff
@@ -98,6 +50,9 @@ steps:
             filename: start.bat
             type: TEXT
             data: |
+              echo "DEBUG: Path mapping rules file: {{Session.PathMappingRulesFile}}"
```
leongdl: nit; cleanup?

src/deadline/max_submitter/default_max_job_template.yaml
```diff
@@ -18,54 +18,6 @@ parameterDefinitions:
       patterns:
       - '*'
   description: The Max scene file to render.
-- name: Frames
```
leongdl, line 21: If we got rid of all these settings, are they still mutable for job bundles?

jericht: They are. They are adding dynamically now instead of being a single parameter here. I can paste example job bundles in both modes.

src/deadline/max_submitter/utilities/max_utils.py
jericht: Yes, this is following what D10 did. There's probably another way to recover the original scene settings without hold and fetch (it's really slow) but that is out of scope here.

src/deadline/max_adaptor/MaxClient/render_handlers/vray_handler.py
jericht: Yes, I've changed the subclasses to not override `start_render`. In a previous iteration, the code was using `rt.render`'s `outputFileName` parameter (or something like that) to set the output file, but it was since changed to just setting `rt.rendOutputFilename`. Maybe this can be undone.

src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
jericht: I've deleted it, this was debug code that revealed an issue in import order when using override adaptors dev feature. The system install of the adaptor was taking precedence over the override when importing modules.

src/deadline/max_adaptor/MaxAdaptor/adaptor.py
jericht: That's a great idea, but can be done in a separate PR. This one is already too large.

src/deadline/max_adaptor/MaxAdaptor/adaptor.py
jericht: Yeah that's fine. It will just skip printing the log in that case. We can accept feature request to make a configurable install path or figure it out somehow in a future revision.

src/deadline/max_submitter/default_max_job_template.yaml
jericht: Thanks, I've deleted these debug lines

src/deadline/max_submitter/default_max_job_template.yaml
jericht: I've updated the PR description with example job bundles from both submission modes.

seant-aws: just a callout, we're also talking in person.

With what we experienced recently with the DCC canary tests, we need to make sure breaking changes also keep those tests in mind.

Our plan is to pin the deadline-cloud-for-3ds-max version in the hostconfig for the DCC Canary tests, so that when we do eventually release, we don't break the tests all over again

src/deadline/max_submitter/max_render_submitter.py
```diff
@@ -100,67 +155,39 @@ def on_create_job_bundle_callback(
...
+            for state_set in state_sets_to_submit:
+                state_set.frame_range = settings.frame_list
 
-    # Add render element output directories to output_directories set
-    if settings.render_elements and not settings.ignore_render_elements_by_name:
-        try:
-            render_element_dirs = get_render_elements_output_directories()
-            output_directories.update(render_element_dirs)
-            _logger.debug(f"Added render element output directories: {render_element_dirs}")
+        # Add render element output directories to output_directories set
```
jericht, line 175: Move this outside of the state set if block

src/deadline/max_adaptor/MaxAdaptor/adaptor.py
```diff
@@ -312,7 +322,8 @@ def _populate_action_queue(self) -> None:
         )
 
         for action_name in _FIRST_MAX_ACTIONS:
-            self._action_queue.enqueue_action(self._action_from_action_item(action_name))
+            if action_name in self.init_data:
```
jericht, line 333: Consider fixing this breaking change

jericht:
Note render elements may not be working with batch render jobs at the moment. This can be addressed in this PR or in a future PR.

src/deadline/max_adaptor/MaxAdaptor/adaptor.py
```diff
@@ -288,7 +298,7 @@ def _start_max_client(self) -> None:
         python_path_addition = f"{openjd_namespace_dir}{os.pathsep}{deadline_namespace_dir}"
         if "PYTHONPATH" in os.environ:
             os.environ["PYTHONPATH"] = (
-                f"{os.environ['PYTHONPATH']}{os.pathsep}{python_path_addition}"
+                f"{python_path_addition}{os.pathsep}{os.environ['PYTHONPATH']}"
```
leongdl, line 304: Why swap this around? Would we run the wrong python here? This is stating 3dsmax.exe below, can the python path start the wrong pythong (eg not the 3dsmax one?

Can we add a comment why?

src/deadline/max_adaptor/MaxAdaptor/adaptor.py
leongdl: So we are only running max actions that are in the init data as a guardrail?

src/deadline/max_adaptor/MaxAdaptor/adaptor.py
leongdl: yeah separate, lets not forget this.

src/deadline/max_adaptor/MaxClient/render_handlers/vray_handler.py
leongdl: Thats fine, as long as the code flow is clear.

leongdl (dismissed):
I think we need to merge this version, and then do small incremental improvement PRs. 

There is probably a follow up for integ test already.

src/deadline/max_adaptor/MaxAdaptor/adaptor.py
jericht: We need to run 3dsMax's Python, not the system Python. I discovered an issue where the adaptor code was importing modules from the system Python site-packages rather than from 3dsMax's site-packages.

Putting the `python_path_addition` at the end means that if a dependency is found in an earlier Python site-packages in `PYTHONPATH`, it will be prioritized. We don't want that here.

I will add a comment

seant-aws:
I ran it locally and played around with the batch rendering. Seems good. I like the UX for new button opening the batch rendering menu. Job bundles exporting seems good, and the I've used the adaptor test script to run the sample bundle and do a batch render locally. Feature seems to be working overall. It also seems like render elements may be working? there seems to be output for the batch I did.
Looks like we already know next steps, so this can be a good base feature for batch render.

```
Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----         3/19/2026   4:48 PM        1094195 new_View02_Diffuse.png
-a----         3/19/2026   4:48 PM          86504 new_View02_Alpha.png
-a----         3/19/2026   4:48 PM        2293175 new_View02.png
-a----         3/19/2026   4:45 PM           7272 new_View01_Diffuse.png
-a----         3/19/2026   4:45 PM           7272 new_View01_Alpha.png
-a----         3/19/2026   4:45 PM        1079492 new_View01.png
```

src/deadline/max_adaptor/MaxAdaptor/adaptor.py
```diff
@@ -97,7 +107,7 @@ class MaxAdaptor(Adaptor[AdaptorConfiguration]):
 
     @property
     def integration_data_interface_version(self) -> SemanticVersion:
-        return SemanticVersion(major=0, minor=1)
+        return SemanticVersion(major=0, minor=3)
```
jericht, line 110: Change

src/deadline/max_adaptor/MaxAdaptor/adaptor.py
```diff
@@ -343,6 +362,53 @@ def _populate_action_queue(self) -> None:
...
+
+        MaxScript's log buffer truncates output at 512 characters per line, so the
+        STDOUT captured by the adaptor may be incomplete. The on-disk Max.log contains
+        the untruncated output. This method is called on failure so the full error
+        context is available in the worker logs.
+
+        Lines longer than _MAX_LOG_LINE_LENGTH are split into chunks to avoid any
+        downstream truncation in log pipelines.
+        """
+        import glob
```
jericht, line 377: move to top

src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
```diff
@@ -145,6 +164,22 @@ def start_render(self, data: dict) -> None:
             )
 
             output_name = self.reformat_framenumber_padding(output_name, frame)
+
+            # Let subclasses configure renderer-specific settings with the resolved name
+            self.log_to_console(
+                f"DEBUG start_render: output_dir='{self.output_dir}', output_name='{output_name}', output_format='{self.output_format}'"
+            )
```
jericht, line 171: remove

seant-aws (approved):

src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
```diff
@@ -300,6 +358,9 @@ def set_output_file_path(self, data: dict) -> None:
         """
         Sets the output file path.
 
+        Note: Path mapping is already applied by Deadline Cloud to job parameter values
+        before they reach the adaptor, so no additional mapping is needed here.
+
```
jericht, line 363: remove

leongdl (approved):
Lets push this, fix comments with smaller PRs, polish it a few rounds, get the integ tests going, add more test and release!

src/deadline/max_adaptor/MaxClient/render_handlers/default_max_handler.py
```diff
@@ -377,9 +438,79 @@ def set_state_set(self, data: dict) -> None:
...
+        else:
+            self.log_to_console(f"Set state set to: {state_set_name}")
 
         self.check_renderer()
 
+    def set_scene_state(self, data: dict) -> None:
+        """
+        Restore a scene state via rt.sceneStateMgr.
+
+        This is distinct from set_state_set which uses the Autodesk.Max.StateSets.Plugin API.
```
jericht, line 450: Verify this is a real API name, otherwise just use plain English to describe the API

src/deadline/max_submitter/sanity_checks.py
```diff
@@ -87,6 +101,16 @@ def check_sanity(settings: RenderSubmitterUISettings):
     if not settings.name:
         raise Exception("No Job Name was given")
 
+    # Warn if the output filename pattern has no frame padding token (default mode only)
+    if settings.submission_mode == SubmissionMode.DEFAULT.value:
+        if "#" not in settings.output_filename_pattern:
+            warnings.append(
+                "The output filename pattern does not contain a frame padding token (#). "
+                "Each rendered frame will overwrite the previous one."
+            )
```
jericht, line 110: A similar warning should apply for batch render filenames that do not have the frame token. These batch render output filenames are queried from the batch views.

test/unit/deadline_adaptor_for_3ds_max/MaxAdaptor/test_adaptor.py
```diff
@@ -53,7 +56,7 @@
 
 # Expected version - must be bumped when schemas change
 EXPECTED_SCHEMA_VERSION_MAJOR = 0
-EXPECTED_SCHEMA_VERSION_MINOR = 1
+EXPECTED_SCHEMA_VERSION_MINOR = 3
```
jericht, line 59: Update this as well

test/unit/deadline_adaptor_for_3ds_max/MaxAdaptor/test_adaptor.py
```diff
@@ -92,7 +95,7 @@ class TestMaxAdaptor_semantic_version:
     def test_semantic_version(self, init_data: dict) -> None:
         """Tests that the adaptor semantic version is in the expected format"""
         adaptor = MaxAdaptor(init_data)
-        assert adaptor.integration_data_interface_version == SemanticVersion(major=0, minor=1)
+        assert adaptor.integration_data_interface_version == SemanticVersion(major=0, minor=3)
```
jericht, line 98: Same here

test/unit/deadline_adaptor_for_3ds_max/MaxClient/render_handlers/test_max_handler_base.py
```diff
@@ -164,3 +164,91 @@ def test_cleanup_render_elements_success(self, maxhandlerbase: DefaultMaxHandler
 
         # THEN
         mock_render_element_manager.restore_render_elements.assert_called_once()
+
+
+class TestBatchRender:
```
jericht, line 169: These unit tests seem incomplete

test/unit/deadline_shared_for_3ds_max/test_shared_max_utils.py
```diff
@@ -872,3 +797,235 @@ def test_configure_vray_raw_output_handles_exception(mock_rt: MagicMock) -> None
...
+            "missing_frame_end",
+            "missing_width",
+            "missing_height",
+        ],
+    )
+    def test_has_all_overrides(
+        self, override_preset, frame_start, frame_end, width, height, expected
+    ):
+        """Verify has_all_overrides property correctly identifies complete overrides."""
+        from deadline.max_shared.utilities.max_utils import BatchRenderView
```
jericht, line 828: move to top

same with all other function level imports that do not need to be imported at the function level

test/unit/deadline_shared_for_3ds_max/test_shared_max_utils.py
```diff
@@ -872,3 +797,235 @@ def test_configure_vray_raw_output_handles_exception(mock_rt: MagicMock) -> None
...
+        result = _view_to_batch_render_view(mock_view, 1)
+
+        assert result.frame_start is None
+        assert result.frame_end is None
+        assert result.width is None
+        assert result.height is None
+        assert result.pixel_aspect is None
+
+
+class TestGetbatchRenderviews:
```
jericht, line 969: Make this proper PascalCase

test/unit/deadline_shared_for_3ds_max/test_shared_max_utils.py
```diff
@@ -872,3 +797,235 @@ def test_configure_vray_raw_output_handles_exception(mock_rt: MagicMock) -> None
...
+    """Tests for get_batch_render_views function."""
+
+    @pytest.fixture(autouse=True)
+    def mock_rt(self):
+        """Mock the pymxs runtime for all tests in this class."""
+        with patch("deadline.max_shared.utilities.max_utils.rt") as mock:
+            mock.undefined = object()
+            yield mock
+
```
jericht, line 978: This can be a module-level autouse mock, we never want to actually call Max APIs in unit tests

test/unit/deadline_submitter_for_3ds_max/test_create_job_bundle.py
```diff
@@ -0,0 +1,784 @@
...
+    def test_creates_step_params_without_batch(
+        self,
+        mock_get_render_elements,
+        mock_get_batch_views,
+        default_job_template,
+        mock_settings,
+        sample_state_set,
+    ):
+        """Verify step-specific parameters are created without batch rendering."""
+        from deadline.max_submitter.create_job_bundle import _create_param_definitions
```
jericht, line 98: to top

move all function level imports in test function to be top module level imports for all test files in this PR

test/unit/deadline_submitter_for_3ds_max/test_data_classes.py
```diff
@@ -39,3 +46,168 @@ def test_no_last_rend_output_filename_field(self) -> None:
...
+            state_set=sample_state_set,
+            batch_view=None,
+            frame_range="1-100",
+            width=1920,
+            height=1080,
+        )
+
+        assert step_data.name == "MyStateSet"
+
+    def test_name_with_batch_view(self, sample_state_set):
```
jericht, line 81: Add error case test no state set or batch

PR #207, opened 2026-03-06
chore: fix integ test to work with current code

AlexTranAmz:
Integ test are currently broken due to multiple reasons

Fix integ test by:
- Add code to properly add RE to name when render
- Export env var correctly to hatch env
- Decode log and change how path mapping test looking for log to remove flakiness
- Add xfail to vray re test due to 3dsmax crash on weaker machine
- Change scene file to support 2024->2026 and output file name to follow our new code changes
- Fix some job template to also follow this new file name logic

Integ test now passing again

Integ test on all 3 version of 3dsMax passing

Yes

- [ ] Yes
- [x] No

   - [ ] Yes
   - [x] No

No

----


test/integ/conftest.py
```diff
@@ -1,8 +1,75 @@
...
+        with winreg.OpenKey(
+            winreg.HKEY_LOCAL_MACHINE,
+            r"SYSTEM\CurrentControlSet\Control\Session Manager\Environment",
+        ) as key:
+            i = 0
+            while True:
+                try:
+                    name, value, _ = winreg.EnumValue(key, i)
+                    if name not in os.environ:
+                        os.environ[name] = value
```
epmog, line 37: I'd presume that our usage of this is probably pretty small (adding new envs). Do we need to update existing ones?

test/integ/conftest.py
```diff
@@ -1,8 +1,75 @@
...
+    installer) live in the registry. A running process only sees them if it
+    was started AFTER they were written, AND the parent process propagated
+    them. In CI (CodeBuild), the agent process may have been started before
+    the installer ran, so its children never inherit the new vars even though
+    they exist in the registry.
+    """
+    if os.name != "nt":
+        return
+
+    import winreg
```
seant-aws, line 25: tiny nit: should the import be on top

epmog (approved):

test/integ/conftest.py
AlexTranAmz: This is needed due to sometime env var in the system is not being imported correctly to the hatch env. That's the main issue rather than just adding new envs

test/integ/conftest.py
AlexTranAmz: discuss offline, not relevant for this instance

seant-aws (approved):

PR #212, opened 2026-03-10
chore: change checksum to be lower cases to match actual

AlexTranAmz:
Checksum was accidentally all caps lock which is not what we expected

Lowercase all the checksum string

Integ test setup will run as expected

Local deployment and run test against mainline

No

- [ ] Yes
- [x] No

   - [ ] Yes
   - [x] No

No

----


leongdl (approved):

pipeline/setup-runner.py
epmog, line 86: should we upper/lower here so that it doesn't matter how the user enters it? Maybe makes the manual change a little bit easier?

for the hash, the upper/lower doesn't change the value of the hash so we should be good

epmog (approved):
lower is default for powershell command. approving

pipeline/setup-runner.py
AlexTranAmz: Check and confirm we generate lower case by default, this is a format issue on the previous PR

PR #217, opened 2026-03-13
feat: add V-Ray export workflow with tile rendering

ZainAallii:

3ds Max users with V-Ray needed the ability to export .vrscene files and render them using V-Ray Standalone on Deadline Cloud, with support for tile-based region rendering to handle high-resolution outputs. The existing submitter only supported direct 3ds Max rendering through the adaptor.

Added a V-Ray Standalone export workflow with two modes:
- **Local export**: Export .vrscene on the workstation, then render on the farm
- **Farm export**: Export and render entirely on the farm using 3dsmaxcmd + V-Ray Standalone

Tile rendering splits frames into an N×M grid using V-Ray's ``-region`` flag, renders each tile as a separate task, then merges them using additive compositing (Pillow for PNG/TIFF/JPG, OpenEXR+numpy for EXR). A V-Ray Export tab is injected into the existing submitter when V-Ray is the active renderer.

Key files: submitter logic, job template generation with embedded render/merge scripts, UI tab, settings dataclass, workflow registry, executable path discovery, and OS-agnostic path mapping.

- Adds a new V-Ray Standalone submission workflow (no changes to existing 3ds Max adaptor workflow)
- Modifies ``submit_dialog.py`` to inject V-Ray Export tab (only when V-Ray is detected)
- Adds one constant to ``data_const.py``
- No changes to adaptor code

- 52 new unit tests covering tile coordinate calculation, frame range parsing, output format validation, and vrscene filename generation (all passing)
- Full test suite: 258 passed, 0 failed
- Lint (ruff) and format (black) checks passing
- Manual end-to-end testing: local export + render, farm export + render, tile rendering with merge, on Deadline Cloud farm with 3ds Max 2025 + V-Ray 6

Yes. Architecture doc at ``docs/design/vray-standalone-file-map.md`` and design specs under ``.kiro/specs/``.

- [ ] Yes
- [x] No

No. This is a new feature with no changes to existing workflows or APIs.

----

[test-results.txt](https://github.com/user-attachments/files/25977890/test-results.txt)
[fmt-results.txt](https://github.com/user-attachments/files/25977922/fmt-results.txt)
[lint-results.txt](https://github.com/user-attachments/files/25977926/lint-results.txt)

### DIFF
--- a/docs/design/vray-standalone-file-map.md
+++ b/docs/design/vray-standalone-file-map.md
@@ -1,0 +1,111 @@
+# V-Ray Standalone Submitter — File Map
+
+Overview of all files added for the V-Ray Standalone workflow.
+
+```
+src/deadline/max_submitter/
+├── vray_standalone_submitter.py        — Main submitter logic: job bundle creation for
+│                                         local export and farm export modes, path mapping
+│                                         script loading, vrscene path fix script loading,
+│                                         tile rendering integration
+├── vrscene_settings.py                 — Settings dataclass with sticky persistence (export
+│                                         mode, region grid, output format, movie settings)
+├── ui/
+│   └── vray_standalone_tab.py          — Qt UI tab: export mode radios, job options, region
+│                                         grid spinners, output format dropdown, browse button
+├── utilities/
+│   ├── vrscene_job_submission.py       — Job template loading and parameter builders:
+│   │                                     tile coordinate calculation, script injection,
+│   │                                     parameter value builders for all job types
+│   ├── vrscene_utils.py                — V-Ray helpers: renderer detection, vrscene export
+│   │                                     (local), settings validation, output format detection
+│   └── vray_executable_utils.py        — Executable path resolution for vray.exe and
+│                                         3dsmaxcmd.exe (env vars + install path fallback)
+├── scripts/
+│   ├── fix_vrscene_paths.py            — Farm script: runs 3dsmaxcmd export then reverses
+│   │                                     session paths so render -remapPath works correctly
+│   ├── path_mapping_render.py          — Farm script: reads path mapping rules and calls
+│   │                                     vray.exe with -remapPath arguments
+│   ├── tile_render.py                  — Farm script: renders a single tile region with vray.exe
+│   ├── tile_merge.py                   — Farm script: merges tile images into complete frames
+│   │                                     using Pillow (PNG/JPEG/TIFF) or OpenEXR
+│   ├── create_movie.py                 — Farm script: creates MP4 from frames (future release)
+│   └── export_vrscene_farm.ms          — MAXScript: exports vrscene on the farm worker via
+│                                         3dsmaxcmd, called by fix_vrscene_paths.py
+└── job_templates/
+    ├── vray_render_job_template.yaml   — OpenJD template: single-step vray.exe render
+    ├── vray_export_job_template.yaml   — OpenJD template: vrscene export step (farm mode)
+    ├── vray_tile_render_job_template.yaml — OpenJD template: RenderRegions + MergeRegions steps
+    └── vray_combined_job_template.yaml — OpenJD template: export + render combined (farm mode)
+
+install_files/
+└── AWSDeadline-SubmitToDeadlineCloud-VRayStandalone.mcr  — 3ds Max macro for menu entry
+                                                             (installer bundling: follow-up PR)
+
+force_reload.ms                         — Dev helper: reloads Python modules in 3ds Max
+```
+
+## Architecture Integration
+
+How the V-Ray Standalone feature relates to the existing submitter/adaptor architecture.
+
+### Existing Architecture Flow
+
+```
+run_ui.py
+  → show_job_bundle_submitter()
+    → SubmitMaxJobToDeadlineDialog(
+        job_setup_widget_type = SceneSettingsWidget,
+        on_create_job_bundle_callback = on_create_job_bundle_callback
+      )
+```
+
+The existing callback builds an OpenJD job template that uses the `3dsmax-openjd` adaptor.
+The farm worker launches 3ds Max and renders inside it.
+
+### V-Ray Hook Points
+
+The V-Ray feature hooks into the existing architecture at exactly two points:
+
+1. **`submit_dialog.py`** — `SubmitMaxJobToDeadlineDialog.__init__()` intercepts the original
+   callback, wraps it in `_on_create_job_bundle_wrapper()`, and calls
+   `_add_vray_export_tab_if_available()`. When V-Ray is the active renderer, a "V-Ray Export"
+   tab appears in the dialog.
+
+2. **Submit-time routing** — The wrapper checks `vray_export_widget.is_vray_export_enabled()`:
+   - Enabled → routes to `on_create_vrscene_job_bundle_callback` with `VRSceneRenderSubmitterUISettings`
+   - Disabled → passes through to the original `on_create_job_bundle_callback` with `RenderSubmitterUISettings`
+
+No other existing files are modified (except `data_const.py` for the settings file extension constant).
+
+### Adaptor vs Adaptor-Free
+
+| | Existing 3ds Max Workflow | V-Ray Standalone Workflow |
+|---|---|---|
+| Rendering | `3dsmax-openjd` adaptor launches 3ds Max on the farm | Farm scripts call `vray.exe` directly |
+| Job template | Loaded from `default_max_job_template.yaml` | Loaded from YAML files in `job_templates/` |
+| Path mapping | Handled by the adaptor | `path_mapping_render.py` reads `Session.PathMappingRulesFile` |
+| Settings class | `RenderSubmitterUISettings` | `VRSceneRenderSubmitterUISettings` |
+| UI tab | `SceneSettingsWidget` | `VRayStandaloneSettingsWidget` |
+| Sticky settings ext | `.deadline_render_settings.json` | `.deadline_vrscene_settings.json` |
+
+### Executable Path Resolution
+
+Farm workers need `vray.exe` (Windows) on the `PATH` or via environment variable.
+Set `VRAY_EXECUTABLE` to the full path of `vray.exe` on each worker — this is the recommended
+approach and should be configured in the fleet host configuration or worker environment.
+The fallback derives the path from `VRAY_FOR_3DSMAX{version}_MAIN` (set by the V-Ray installer).
+
+Similarly, farm export mode requires `3dsmaxcmd.exe`. Set `MAXCMD_EXECUTABLE` to its full path,
+or the submitter will derive it from the current 3ds Max installation at submit time.
+
+### Coupling Summary
+
+The V-Ray feature is a parallel workflow that shares the dialog shell but has its own:
+- Settings dataclass (`vrscene_settings.py`)
+- UI tab (`vray_standalone_tab.py`)
+- Submit callback (`vray_standalone_submitter.py`)
+- Job templates (`job_templates/`)
+- Farm scripts (`scripts/`)
+
+Everything is additive. The only modified existing files are `submit_dialog.py` (callback wrapper + tab injection) and `data_const.py` (one constant).
--- a/install_files/AWSDeadline-SubmitToDeadlineCloud-VRayStandalone.mcr
+++ b/install_files/AWSDeadline-SubmitToDeadlineCloud-VRayStandalone.mcr
@@ -1,0 +1,49 @@
+-- AWS Deadline Cloud - V-Ray Standalone Submitter Macro
+-- Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+macroScript SubmitToDeadlineCloudVRayStandalone
+	category:"AWS Deadline Cloud"
+	buttonText:"Submit V-Ray Standalone"
+	toolTip:"Submit V-Ray Standalone Render Job To Deadline Cloud"
+	icon:#("Deadline",1)
+(
+	-- Check if V-Ray is the current renderer
+	local isVRay = false
+	try
+	(
+		local rendererName = (classof renderers.current) as string
+		isVRay = (findString rendererName "VRay" != undefined) or (findString rendererName "V_Ray" != undefined)
+	)
+	catch
+	(
+		isVRay = false
+	)
+	
+	if not isVRay then
+	(
+		messageBox "V-Ray Standalone workflow requires V-Ray as the active renderer.\n\nPlease set V-Ray as your current renderer to use this workflow." title:"V-Ray Not Active"
+		return false
+	)
+	
+	-- Import the Python module and run the UI
+	python.Execute "import sys"
+	python.Execute "import os"
+	
+	-- Try development path first, then installed path
+	local devPath = "C:\\Users\\Administrator\\gitrepos\\deadline-cloud-for-3ds-max\\src\\deadline\\max_submitter"
+	local installedPath = (getDir #scripts) + "\\deadline\\max_submitter"
+	
+	local submitterPath = installedPath
+	if (doesFileExist (devPath + "\\run_vray_standalone_ui.py")) then (
+		submitterPath = devPath
+		format "Using development path: %\n" submitterPath
+	) else (
+		format "Using installed path: %\n" submitterPath
+	)
+	
+	python.Execute ("sys.path.insert(0, r'" + submitterPath + "')")
+	
+	-- Launch the V-Ray Standalone submitter
+	python.Execute "from run_vray_standalone_ui import main"
+	python.Execute "main()"
+)
--- a/scripts/MAXScript_VRayStandaloneExport.ms
+++ b/scripts/MAXScript_VRayStandaloneExport.ms
@@ -1,0 +1,104 @@
+-- V-Ray Standalone Export MAXScript Job
+-- Runs on Deadline workers to export vrscene files
+-- Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+try
+(
+	local du = DeadlineUtil  -- Interface provided by Deadline 3ds Max plugin
+	if du == undefined do
+	(
+		-- Stand-in struct for testing outside Deadline
+		struct DeadlineUtilStruct
+		(
+			fn SetTitle title = ( format "Title: %\n" title ),
+			fn SetProgress percent = (true),
+			fn FailRender msg = ( throw msg ),
+			fn GetJobInfoEntry key = ( undefined ),
+			fn LogMessage msg = ( format "INFO: %\n" msg ),
+			fn WarnMessage msg = ( format "WARNING: %\n" msg ),
+			CurrentFrame = ((sliderTime as string) as integer),
+			CurrentTask = ( -1 ),
+			SceneFileName = ( maxFilePath + maxFileName ),
+			SceneFilePath = ( maxFilePath )
+		)
+		du = DeadlineUtilStruct()
+	)
+	
+	fn ExportToVRayStd exportPath startFrame endFrame incrBaseFrame: =
+	(
+		-- Create output directory
+		makeDir (getFileNamePath exportPath) all:true
+		sysinfo.currentdir = getFileNamePath exportPath
+		
+		-- Ensure VRayRT is the active shade renderer
+		if not (iskindof renderers.activeShade VRayRT) do 
+			renderers.activeShade = VRayRT()
+		
+		-- Export based on frame range
+		if startFrame == endFrame then
+			-- Single frame export
+			vrayExportRTScene exportPath separateFiles:false
+		else if incrBaseFrame != unsupplied and incrBaseFrame != startFrame then
+			-- Incremental export (not first frame)
+			vrayExportRTScene exportPath separateFiles:false startFrame:startFrame endFrame:endFrame incrBaseFrame:incrBaseFrame
+		else
+			-- Full animation or first frame of incremental
+			vrayExportRTScene exportPath separateFiles:false startFrame:startFrame endFrame:endFrame
+	)
+	
+	local st = timestamp()
+	
+	du.SetTitle "VRSCENE Export MAXScript Job"
+	du.LogMessage ">Starting VRSCENE Export MAXScript Job..."
+	
+	-- Get job parameters
+	local startFrame = (du.GetJobInfoEntry "Render_StartFrame") as integer
+	local endFrame = (du.GetJobInfoEntry "Render_EndFrame") as integer
+	local vrsceneName = du.GetJobInfoEntry "Render_InputFilename"
+	local exportAnimationMode = du.GetJobInfoEntry "Render_ExportAnimationMode"
+	
+	-- Default to Single File mode
+	if exportAnimationMode == "" do
+		exportAnimationMode = "1"
+	exportAnimationMode = exportAnimationMode as integer
+	
+	du.LogMessage ">Exporting VRSCENE File..."
+	
+	case exportAnimationMode of
+	(
+		1: (
+			-- Single File: Export entire animation to one vrscene
+			ExportToVRayStd vrsceneName startFrame endFrame
+		)
+		2: (
+			-- File Per Frame: Export current frame to separate vrscene
+			local frameVrsceneName = (getFilenamePath vrsceneName) + (getFilenameFile vrsceneName) + "." + (formattedPrint du.CurrentFrame format:"04i") + (getFilenameType vrsceneName)
+			ExportToVRayStd frameVrsceneName du.CurrentFrame (du.CurrentFrame + 1)
+		)
+		3: (
+			-- File Per Frame (Incremental): Export with reference to first frame
+			local frameVrsceneName = (getFilenamePath vrsceneName) + (getFilenameFile vrsceneName) + "." + (formattedPrint du.CurrentFrame format:"04i") + (getFilenameType vrsceneName)
+			ExportToVRayStd frameVrsceneName du.CurrentFrame (du.CurrentFrame + 1) incrBaseFrame:startFrame
+		)
+	)
+	
+	du.LogMessage ("+Finished VRSCENE Export MAXScript Job in "+ ((timestamp() - st)/1000.0) as string + " sec.")
+	true
+)
+catch
+(
+	-- Error handling with stack trace (Max 2017+)
+	if ((maxVersion())[1]/1000 as integer) >= 19 then
+	(
+		if hasCurrentExceptionStackTrace() then
+		(
+			local errorMessage = getCurrentException()
+			DeadlineUtil.WarnMessage("MaxScript Error: " + errorMessage)
+			local stackTrace = getCurrentExceptionStackTrace()
+			stackTrace = filterString stackTrace "\n"
+			for line in stackTrace do
+				DeadlineUtil.WarnMessage(line)
+		)
+	)
+	throw()
+)
--- a/src/deadline/max_submitter/data_const.py
+++ b/src/deadline/max_submitter/data_const.py
@@ -109,3 +109,17 @@ STEREO_CAMERA_OPTIONS = [
     ["Left, Right and Center", "All"],
     ["Disable Stereo Camera Submission", "None"],
 ]
+
+# V-Ray Standalone Workflow Constants
+VRSCENE_EXPORT_MODES = [
+    ["Export VRSCENE On This Workstation", 1],
+    ["Export VRSCENE On Deadline", 2],
+]
+
+VRSCENE_EXPORT_ANIMATION_MODES = [
+    ["Single File", 1],
+    ["File Per Frame", 2],
+    ["File Per Frame (Incremental)", 3],
+]
+
+VRSCENE_SUBMITTER_SETTINGS_FILE_EXT = ".deadline_vrscene_settings.json"
--- a/src/deadline/max_submitter/run_vray_standalone_ui.py
+++ b/src/deadline/max_submitter/run_vray_standalone_ui.py
@@ -1,0 +1,38 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
+Entry point for V-Ray Standalone submitter UI.
+This can be called from MAXScript or directly from Python.
+"""
+
+import logging
+
+# Configure logging
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s [%(levelname)s] %(name)s: %(message)s",
+)
+
+_logger = logging.getLogger(__name__)
+
+
+def main():
+    """
+    Main entry point for V-Ray Standalone submitter.
+    """
+    try:
+        from vray_standalone_submitter import show_vray_standalone_submitter
+
+        _logger.info("Launching V-Ray Standalone submitter...")
+        window = show_vray_standalone_submitter()
+        return window
+    except Exception as e:
+        _logger.error(f"Failed to launch V-Ray Standalone submitter: {e}")
+        import traceback
+
+        traceback.print_exc()
+        raise
+
+
+if __name__ == "__main__":
+    main()
--- a/src/deadline/max_submitter/scripts/create_movie.py
+++ b/src/deadline/max_submitter/scripts/create_movie.py
@@ -1,0 +1,68 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#!/usr/bin/env python3
+# Create Movie Script
+# Creates an MP4 movie from rendered frames using ffmpeg.
+
+import os
+import subprocess
+import sys
+
+
+def normalize_path(p):
+    # Convert backslashes to forward slashes to avoid escape issues.
+    return p.replace("\\", "/")
+
+
+def main():
+    create_movie = r"{{Param.CreateMovie}}"
+    if create_movie != "true":
+        print("Movie creation disabled, skipping")
+        return 0
+
+    output_dir = normalize_path(r"{{Param.OutputDir}}")
+    output_filename = r"{{Param.OutputFileName}}"
+    movie_filename = r"{{Param.MovieFilename}}"
+    frame_rate = int(r"{{Param.FrameRate}}")
+
+    base, ext = os.path.splitext(output_filename)
+
+    # ffmpeg input pattern: base.%04d.ext
+    input_pattern = os.path.join(output_dir, f"{base}.%04d{ext}")
+    movie_path = os.path.join(output_dir, movie_filename)
+
+    print("=== Create Movie ===")
+    print(f"Input pattern: {input_pattern}")
+    print(f"Output: {movie_path}")
+    print(f"Frame rate: {frame_rate}")
+
+    cmd = [
+        "ffmpeg",
+        "-framerate",
+        str(frame_rate),
+        "-i",
+        input_pattern,
+        "-c:v",
+        "libx264",
+        "-pix_fmt",
+        "yuv420p",
+        "-crf",
+        "18",
+        movie_path,
+    ]
+
+    print(f"Running: {' '.join(cmd)}")
+
+    try:
+        subprocess.run(cmd, check=True)
+        print(f"Movie created: {movie_path}")
+        return 0
+    except FileNotFoundError:
+        print("ERROR: ffmpeg not found. Cannot create movie.", file=sys.stderr)
+        return 1
+    except subprocess.CalledProcessError as e:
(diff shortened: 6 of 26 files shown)


.kiro/specs/vray-path-mapping/design.md
```diff
@@ -0,0 +1,468 @@
+# V-Ray Path Mapping Design
```
leongdl: Please don't check in the specs :) We may want to archive them somewhere, like on forums.

docs/design/vray-standalone-file-map.md
```diff
@@ -0,0 +1,108 @@
+# V-Ray Standalone Submitter — File Map
+
+Overview of all files added for the V-Ray Standalone workflow.
```
leongdl, line 3: Please do keep this file!

install_files/AWSDeadline-SubmitToDeadlineCloud-VRayStandalone.mcr
```diff
@@ -0,0 +1,49 @@
...
+		messageBox "V-Ray Standalone workflow requires V-Ray as the active renderer.\n\nPlease set V-Ray as your current renderer to use this workflow." title:"V-Ray Not Active"
+		return false
+	)
+	
+	-- Import the Python module and run the UI
+	python.Execute "import sys"
+	python.Execute "import os"
+	
+	-- Try development path first, then installed path
+	local devPath = "C:\\Users\\Administrator\\gitrepos\\deadline-cloud-for-3ds-max\\src\\deadline\\max_submitter"
```
leongdl, line 33: Not always true but that is ok given the lines below.

install_files/AWSDeadline-SubmitToDeadlineCloud-VRayStandalone.mcr
```diff
@@ -0,0 +1,49 @@
+-- AWS Deadline Cloud - V-Ray Standalone Submitter Macro
+-- Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+macroScript SubmitToDeadlineCloudVRayStandalone
```
leongdl, line 4: We need to bundle this into the installer right? Otherwise this MCR won't be added to the menu

leongdl: This can be in a follow up PR, and we can show you how to do it.

scripts/MAXScript_VRayStandaloneExport.ms
```diff
@@ -0,0 +1,104 @@
+-- V-Ray Standalone Export MAXScript Job
+-- Runs on Deadline workers to export vrscene files
+-- Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+try
```
leongdl, line 5: Curious, did this come from D10? We are running this max script directly in the submitter?

src/deadline/max_submitter/ui/submit_dialog.py
```diff
@@ -17,8 +17,125 @@ class SubmitMaxJobToDeadlineDialog(SubmitJobToDeadlineDialog):
...
+        """
+        # Check if V-Ray Export tab exists and is enabled
+        if hasattr(self, "vray_export_widget") and self.vray_export_widget.is_vray_export_enabled():
+            _logger.info("V-Ray Export is enabled - creating V-Ray export job")
+
+            # Get V-Ray settings from the tab (creates new settings object)
+            vrscene_settings = self.vray_export_widget.update_settings()
+
+            # Update scene file and vrscene filename from current scene
+            from utilities import max_utils
```
leongdl: We usually have all imports at the top of the file unless it is something that needs to be dynamically imported.

src/deadline/max_submitter/utilities/vray_executable_utils.py
```diff
@@ -0,0 +1,95 @@
...
+
+from pymxs import runtime as rt
+
+_logger = logging.getLogger(__name__)
+
+
+def get_vray_executable_path() -> str:
+    """Find vray.exe — checks VRAY_EXECUTABLE env var, then common install paths."""
+    # Check environment variable first
+    env_path = os.environ.get("VRAY_EXECUTABLE", "")
```
leongdl, line 21: We should strongly prefer this option and update our setup guides to always use this. The fallback detection below is best effort and hard to maintain over time.

src/deadline/max_submitter/utilities/vray_executable_utils.py
```diff
@@ -0,0 +1,95 @@
...
+
+    default_path = "C:/Program Files/Chaos Group/V-Ray/Standalone for x64/bin/vray.exe"
+    _logger.warning(f"V-Ray executable not found, using default: {default_path}")
+    return default_path
+
+
+def get_3dsmax_executable_path() -> str:
+    """Find 3dsmaxcmd.exe — checks MAXCMD_EXECUTABLE env var, then common install paths."""
+    # Check environment variable first
+    env_path = os.environ.get("MAXCMD_EXECUTABLE", "")
```
leongdl, line 49: We should put this into all of our documentation as it is the better thing to do. The fallback below is best effort.

src/deadline/max_submitter/utilities/vrscene_job_submission.py
```diff
@@ -0,0 +1,806 @@
...
+
+
+def get_tile_index(column: int, row: int, total_columns: int) -> int:
+    """Row-major tile index: row * cols + col."""
+    return row * total_columns + column
+
+
+def _get_tile_render_script() -> str:
+    """Embedded Python script for rendering a single tile with V-Ray."""
+    return r"""#!/usr/bin/env python3
```
leongdl: Can we put this as a text file and then we source it? It will help with maintainability over time. It is very hard to correctly edit string escaped text.

src/deadline/max_submitter/utilities/vrscene_job_submission.py
```diff
@@ -0,0 +1,806 @@
...
+
+if __name__ == "__main__":
+    sys.exit(main())
+"""
+
+
+def _get_tile_merge_script() -> str:
+    """Embedded Python script for merging tiles into complete frames."""
+    return r"""#!/usr/bin/env python3
+# V-Ray Tile Merge Script
```
leongdl: Same here

src/deadline/max_submitter/utilities/vrscene_job_submission.py
```diff
@@ -0,0 +1,806 @@
...
+        merge_tiles_pillow(tile_paths, total_cols, total_rows, image_width, image_height, final_output)
+
+    return 0
+
+if __name__ == "__main__":
+    sys.exit(main())
+"""
+
+
+def create_tile_rendering_job_template(
```
leongdl, line 105: This is great, but can we start with a default template and just modify specific elements? See this example:
https://github.com/aws-deadline/deadline-cloud-for-3ds-max/blob/mainline/src/deadline/max_submitter/default_max_job_template.yaml

Loaded here:
https://github.com/aws-deadline/deadline-cloud-for-3ds-max/blob/3025ff0d8fcf6cfa2bb11c34792a3a4360bf2fe6/src/deadline/max_submitter/max_render_submitter.py#L72

src/deadline/max_submitter/utilities/vrscene_job_submission.py
```diff
@@ -0,0 +1,806 @@
...
+        "steps": steps,
+    }
+
+    return template
+
+
+def _get_create_movie_script() -> str:
+    """Embedded Python script for creating MP4 from frames (future release)."""
+    return r"""#!/usr/bin/env python3
+# Create Movie Script
```
leongdl: Same as above, refactor as a standalone script that can be read in without escaped text

src/deadline/max_submitter/utilities/vrscene_job_submission.py
```diff
@@ -0,0 +1,806 @@
...
+        frames = [int(f.strip()) for f in frame_string.split(",")]
+        return min(frames), max(frames)
+    else:
+        frame = int(frame_string.strip())
+        return frame, frame
+
+
+def create_export_job_template() -> Dict[str, Any]:
+    """Create job template for vrscene export job (farm mode)."""
+    template = {
```
leongdl: Same, can we move this to a template that is filled out ?

src/deadline/max_submitter/utilities/vrscene_job_submission.py
```diff
@@ -0,0 +1,806 @@
...
+    )
+
+    if script_path.exists():
+        with open(script_path, "r") as f:
+            return f.read()
+    else:
+        _logger.warning(f"MAXScript file not found at {script_path}, using inline script")
+        # Fallback inline script
+        return """
+-- V-Ray Standalone Export Script
```
leongdl: Same, can we move this to a file that is read in.

src/deadline/max_submitter/utilities/vrscene_utils.py
```diff
@@ -0,0 +1,220 @@
...
+        _logger.info(f"Export mode: {export_animation_mode}, Frames: {start_frame}-{end_frame}")
+
+        # Escape backslashes for MAXScript
+        vrscene_path_escaped = vrscene_path.replace("\\", "\\\\")
+        output_dir_escaped = os.path.dirname(vrscene_path).replace("\\", "\\\\")
+        base_name = os.path.splitext(os.path.basename(vrscene_path))[0]
+
+        # Build MAXScript export command
+        export_script = f"""
+        fn ExportToVRayStd exportPath startFrame endFrame incrBaseFrame: =
```
leongdl: I assume this came from D10

src/deadline/max_submitter/utilities/vrscene_utils.py
```diff
@@ -0,0 +1,220 @@
...
+        return False, error_msg
+
+
+def validate_vrscene_export_settings(settings) -> List[str]:
+    """Validate settings, return list of error messages (empty if valid)."""
+    errors = []
+
+    # Check if V-Ray is the current renderer
+    if not is_vray_renderer():
+        errors.append("V-Ray scene export requires V-Ray as the active renderer")
```
leongdl, line 103: We should return here, since the rest of the code no longer makes sense.

src/deadline/max_submitter/utilities/vrscene_utils.py
```diff
@@ -0,0 +1,220 @@
...
+        return "png"
+
+
+def validate_output_format(format_str: str) -> bool:
+    """Check if format is supported (png, exr, tiff, tif, jpg, jpeg)."""
+    # Normalize first
+    normalized = normalize_output_format(format_str)
+
+    # List of supported formats
+    supported_formats = {"png", "exr", "tiff", "tif", "jpg", "jpeg"}
```
leongdl, line 162: No vrimg?

src/deadline/max_submitter/workflows/__init__.py
```diff
@@ -0,0 +1,47 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""Workflow registry for 3ds Max submitter."""
+
+from dataclasses import dataclass
+from typing import Callable, Dict, List, Optional, Type
+
+
+@dataclass
+class WorkflowDefinition:
```
leongdl: What are these workflows? How is it integrated to the submitter?

leongdl: I don't see these workflow files used else where, how is this integrated?

src/deadline/max_submitter/data_const.py
```diff
@@ -109,3 +109,22 @@
...
+]
+
+VRSCENE_EXPORT_ANIMATION_MODES = [
+    ["Single File", 1],
+    ["File Per Frame", 2],
+    ["File Per Frame (Incremental)", 3],
+]
+
+VRSCENE_WORKFLOW_NAME = "vraystandaloneexport"
+VRSCENE_WORKFLOW_TITLE = "Submit V-RAY STANDALONE Render Job To Deadline"
```
leongdl: Is this a specific and separate menu item outside of our normal 3dsmax ?

src/deadline/max_submitter/run_vray_standalone_ui.py
```diff
@@ -0,0 +1,38 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""
+Entry point for V-Ray Standalone submitter UI.
+This can be called from MAXScript or directly from Python.
+"""
+
+import logging
```
leongdl, line 8: Something like this needs a separate setup for the installer, lets chat if we want to keep it.

src/deadline/max_submitter/vray_standalone_submitter.py
```diff
@@ -0,0 +1,943 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+def _get_fix_vrscene_paths_script() -> str:
+    """
+    Python wrapper that exports vrscene via 3dsmaxcmd, then reverses
+    session-specific paths back to originals so render -remapPath works.
+    """
+    return r"""#!/usr/bin/env python3
```
leongdl: Same comment as others, we may want to move these to stand alone files.

src/deadline/max_submitter/vray_standalone_submitter.py
```diff
@@ -0,0 +1,943 @@
...
+                        "command": "python",
+                        "args": ["{{Task.File.SetupAndRender}}"],
+                    }
+                },
+            },
+        }
+        steps = [export_step, render_step]
+
+    # Build job template
+    job_template = {
```
leongdl: Same as the other modes, we can have a stand alone template and then fill in the blanks instead of a nested dictionary setup

src/deadline/max_submitter/utilities/vrscene_utils.py
ZainAallii: Good question — .vrimg is a special case. It's Chaos's proprietary format tied to the V-Ray Frame Buffer (VFB). [Per Chaos docs](https://docs-chaos.atlassian.net/wiki/spaces/VRAY4TOOLS/pages/117637144/VRImg+to+OpenEXR+Converter), vrimg stores rendered data incrementally (bucket by bucket) in full floating-point with all render elements. It's designed as a render output format for resumable rendering, not as a standard image format.

For our V-Ray Standalone workflow, there are a few complications:

vrimg output is controlled by VFB settings baked into the vrscene, not by a simple file extension swap
Tile merging wouldn't work with vrimg — our Pillow/OpenEXR merge pipeline can't read vrimg files. You'd need Chaos's vrimg2exr converter first
vrimg is typically used for resumable rendering workflows, which is a different use case than what we're targeting here
I'd prefer to skip vrimg support for now and add it as a follow-up if there's demand. If we do add it later, we'd likely need a post-render step that converts vrimg → exr using vrimg2exr before the merge step can run.

docs/design/vray-standalone-file-map.md
ZainAallii: Sounds good!! I made changes to the code on based on your request and updated this file.

install_files/AWSDeadline-SubmitToDeadlineCloud-VRayStandalone.mcr
ZainAallii: Yes the changes should be the part of the installer. I need to learn how to build the installers. Thanks.

.kiro/specs/vray-path-mapping/design.md
ZainAallii: removed the specs

install_files/AWSDeadline-SubmitToDeadlineCloud-VRayStandalone.mcr
ZainAallii: Thanks.

src/deadline/max_submitter/ui/submit_dialog.py
ZainAallii: Updated.

src/deadline/max_submitter/utilities/vray_executable_utils.py
ZainAallii: I updated just to the fallback.

src/deadline/max_submitter/utilities/vrscene_job_submission.py
ZainAallii: Done

src/deadline/max_submitter/utilities/vrscene_job_submission.py
ZainAallii: Done

src/deadline/max_submitter/utilities/vrscene_job_submission.py
ZainAallii: Done

src/deadline/max_submitter/utilities/vrscene_utils.py
ZainAallii: Done.

src/deadline/max_submitter/workflows/__init__.py
ZainAallii: I removed them.

src/deadline/max_submitter/vray_standalone_submitter.py
ZainAallii: Done.

src/deadline/max_submitter/vray_standalone_submitter.py
ZainAallii: Updated.

src/deadline/max_submitter/utilities/vrscene_job_submission.py
ZainAallii: Done.

src/deadline/max_submitter/data_const.py
ZainAallii: Its in the 3dsmax submitter UI as a tab. If V-Ray export checkbox is enabled Max submitter wont be used.

src/deadline/max_submitter/run_vray_standalone_ui.py
ZainAallii: Sure.

src/deadline/max_submitter/utilities/vray_executable_utils.py
ZainAallii: "Done — env var is now the primary path, fallback only uses VRAY_FOR_3DSMAX{version}_MAIN which V-Ray's installer sets automatically. Hardcoded path scanning removed. Setup guide update deferred to follow-up PR."

src/deadline/max_submitter/utilities/vrscene_job_submission.py
ZainAallii: "Done — templates are now loaded from YAML files in job_templates/ and specific elements are modified at runtime (name, parameter defaults, script injection). We have separate YAML files per workflow type rather than one shared base, since the V-Ray standalone structure differs significantly from the 3ds Max adaptor template."

src/deadline/max_submitter/utilities/vrscene_utils.py
ZainAallii: "No, written from scratch for this PR. The vrayExportRTScene MAXScript function is part of V-Ray's public API — that's the standard way to export a vrscene programmatically."

src/deadline/max_submitter/data_const.py
ZainAallii: Updated.

src/deadline/max_submitter/vray_standalone_submitter.py
```diff
@@ -0,0 +1,943 @@
...
+"""
+3ds Max Deadline Cloud Submitter - V-Ray Standalone Workflow
+"""
+
+import logging
+import os
+from pathlib import Path
+from typing import Any, Optional
+
+import pymxs  # noqa
```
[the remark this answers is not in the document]
ZainAallii: "pymxs must be imported to initialize the 3ds Max Python environment. The # noqa comment is intentional."

src/deadline/max_submitter/utilities/vrscene_job_submission.py
```diff
@@ -0,0 +1,238 @@
...
+"""
+
+import logging
+import os
+from pathlib import Path
+from typing import Any, Dict, List, Tuple
+
+import yaml
+
+_logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
ZainAallii: fixed.

src/deadline/max_submitter/run_vray_standalone_ui.py
ZainAallii: I created the installer following the instructions in https://github.com/aws-deadline/deadline-cloud-for-3ds-max/blob/mainline/DEVELOPMENT.md#installer-installation-recommended. It installs okay and submits fine.

docs/design/vray-standalone-file-map.md
```diff
@@ -0,0 +1,111 @@
...
+   - Enabled → routes to `on_create_vrscene_job_bundle_callback` with `VRSceneRenderSubmitterUISettings`
+   - Disabled → passes through to the original `on_create_job_bundle_callback` with `RenderSubmitterUISettings`
+
+No other existing files are modified (except `data_const.py` for the settings file extension constant).
+
+### Adaptor vs Adaptor-Free
+
+| | Existing 3ds Max Workflow | V-Ray Standalone Workflow |
+|---|---|---|
+| Rendering | `3dsmax-openjd` adaptor launches 3ds Max on the farm | Farm scripts call `vray.exe` directly |
```
leongdl, line 85: Nit, this works for vray generally right? EG on linux

src/deadline/max_submitter/scripts/create_movie.py
```diff
@@ -0,0 +1,68 @@
...
+    # ffmpeg input pattern: base.%04d.ext
+    input_pattern = os.path.join(output_dir, f"{base}.%04d{ext}")
+    movie_path = os.path.join(output_dir, movie_filename)
+
+    print("=== Create Movie ===")
+    print(f"Input pattern: {input_pattern}")
+    print(f"Output: {movie_path}")
+    print(f"Frame rate: {frame_rate}")
+
+    cmd = [
```
leongdl, line 38: Question is this always needed? The following h.264 formatting may need some customizations?

src/deadline/max_submitter/scripts/fix_vrscene_paths.py
```diff
@@ -0,0 +1,120 @@
...
+            with open(rules_file, "r") as f:
+                data = json.load(f)
+                rules = data.get("path_mapping_rules", [])
+                for rule in rules:
+                    source = rule.get("source_path", "")
+                    dest = rule.get("destination_path", "")
+                    if source and dest:
+                        # Replace dest paths with source paths in the vrscene
+                        # Use both forward-slash and backslash variants
+                        reverse_map[dest.replace(chr(92), "/")] = source.replace(chr(92), "/")
```
leongdl: 92 was? Maybe a constant for better maintainability. Maybe fix this in a follow up PR.

src/deadline/max_submitter/scripts/fix_vrscene_paths.py
```diff
@@ -0,0 +1,120 @@
...
+# with original source paths so the render step's -remapPath works correctly.
+
+import glob
+import json
+import os
+import subprocess
+import sys
+
+
+def normalize_path(p):
```
leongdl, line 14: Is this script even used?

src/deadline/max_submitter/scripts/tile_merge.py
```diff
@@ -0,0 +1,190 @@
...
+# V-Ray -region outputs full-size images (only region has pixels, rest is black).
+# Uses Pillow for PNG/TIFF/JPG, OpenEXR+numpy for EXR.
+
+import os
+import subprocess
+import sys
+import tempfile
+
+# Create a known install directory for pip packages
+_pip_target = os.path.join(tempfile.gettempdir(), "deadline_pip_packages")
```
leongdl, line 14: I think we want to use Conda here? Or why did we explicitly make ti install?

leongdl: This may come as a surprise that we bypassed conda

src/deadline/max_submitter/utilities/vray_executable_utils.py
```diff
@@ -0,0 +1,65 @@
...
+
+import logging
+import os
+
+from pymxs import runtime as rt
+
+_logger = logging.getLogger(__name__)
+
+
+def get_vray_executable_path() -> str:
```
leongdl, line 13: What about cross OS handling?

I guess it depends on `VRAY_EXECUTABLE` being configured?

src/deadline/max_submitter/utilities/vray_executable_utils.py
```diff
@@ -0,0 +1,65 @@
...
+    env_path = os.environ.get("VRAY_EXECUTABLE", "")
+    if env_path and os.path.exists(env_path):
+        return env_path
+
+    # Fallback: derive from V-Ray's own env vars (set by V-Ray installer)
+    try:
+        max_version = 2000 + (int(rt.maxVersion()[0] / 1000.0) - 2)
+        vray_main = os.environ.get(f"VRAY_FOR_3DSMAX{max_version}_MAIN", "")
+        if vray_main:
+            candidate = os.path.join(vray_main, "vray.exe")
```
leongdl, line 30: What about cross OS rendering?

src/deadline/max_submitter/utilities/vrscene_job_submission.py
```diff
@@ -0,0 +1,235 @@
...
+        "FrameRate": str(settings.vrscene_movie_framerate),
+    }
+    for param in template.get("parameterDefinitions", []):
+        if param["name"] in defaults:
+            param["default"] = defaults[param["name"]]
+
+    return template
+
+
+def _get_create_movie_script() -> str:
```
leongdl, line 148: This is not used?

src/deadline/max_submitter/utilities/vrscene_utils.py
```diff
@@ -0,0 +1,221 @@
...
+        _logger.info(f"Export mode: {export_animation_mode}, Frames: {start_frame}-{end_frame}")
+
+        # Escape backslashes for MAXScript
+        vrscene_path_escaped = vrscene_path.replace("\\", "\\\\")
+        output_dir_escaped = os.path.dirname(vrscene_path).replace("\\", "\\\\")
+        base_name = os.path.splitext(os.path.basename(vrscene_path))[0]
+
+        # Build MAXScript export command
+        export_script = f"""
+        fn ExportToVRayStd exportPath startFrame endFrame incrBaseFrame: =
```
leongdl: Can we move this to a .ms file under scripts?

src/deadline/max_submitter/scripts/tile_merge.py
```diff
@@ -0,0 +1,190 @@
...
+        print("  Full-size tiles — additive compositing")
+    else:
+        print("  Tile-sized tiles — offset pasting")
+
+    canvas = Image.new(mode, (image_width, image_height))
+
+    for row in range(grid_rows):
+        for col in range(grid_cols):
+            tile_index = row * grid_cols + col
+            tile_img = Image.open(tile_paths[tile_index])
```
seant-aws, line 63: do these file opens, also L102, need try+excepts ? in case these opens fail, the whole thing might come crashing down without knowing which caused itt ?

src/deadline/max_submitter/utilities/vrscene_job_submission.py
```diff
@@ -0,0 +1,235 @@
...
+        {"name": "VRSceneOutputPath", "value": vrscene_path},
+        {"name": "StartFrame", "value": str(start_frame)},
+        {"name": "EndFrame", "value": str(end_frame)},
+        {"name": "ExportAnimationMode", "value": str(settings.export_animation_mode)},
+    ]
+
+    return parameters
+
+
+def get_frame_range_from_string(frame_string: str) -> Tuple[int, int]:
```
seant-aws, line 206: I might be wrong, but do we handle multiple ranges for 3dsMax ? such as "1-10,20-30", would this be supported? if so, would we need to add onto this function ?

docs/design/vray-standalone-file-map.md
```diff
@@ -0,0 +1,111 @@
...
+│   ├── vrscene_utils.py                — V-Ray helpers: renderer detection, vrscene export
+│   │                                     (local), settings validation, output format detection
+│   └── vray_executable_utils.py        — Executable path resolution for vray.exe and
+│                                         3dsmaxcmd.exe (env vars + install path fallback)
+├── scripts/
+│   ├── fix_vrscene_paths.py            — Farm script: runs 3dsmaxcmd export then reverses
+│   │                                     session paths so render -remapPath works correctly
+│   ├── path_mapping_render.py          — Farm script: reads path mapping rules and calls
+│   │                                     vray.exe with -remapPath arguments
+│   ├── tile_render.py                  — Farm script: renders a single tile region with vray.exe
```
seant-aws, line 29: something for a future PR maybe, from what I grokked from the code:

seems like for tile renders, we split up the image, but the split up 'tiles' are actually the full resolution of the original image, just with the rest of the image as black. Might be a performance/storage issue, we could cut down the tiles to be actual tiles, rather than full size resolutions. If we have many many frames at super high res, we could cut down storage, and therefore performace, by a factor of N tiles. ?

seant-aws: unless of course there's a reason for this in Vray...

docs/design/vray-standalone-file-map.md
ZainAallii: It will only render on Windows for now. Follow up PR is coming.

ZainAallii: Here's the follow up issue: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/issues/229

src/deadline/max_submitter/scripts/create_movie.py
ZainAallii: "Movie creation is optional — controlled by the 'Create Movie' checkbox in the UI.

The checkbox exists in the code but is hidden — `setVisible(False)`. So it's there but not shown to the user yet.

The UI checkbox exists in `vray_standalone_tab.py` but is currently hidden (`setVisible(False)`). Here's the follow up issue: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/issues/230

src/deadline/max_submitter/scripts/fix_vrscene_paths.py
ZainAallii: Yes, `sys` is used at the bottom — `sys.exit(main())`. The import is needed.

It is used — lines 39 and 40 in `fix_vrscene_paths()`: 

"normalize_path is used in fix_vrscene_paths() on lines 39-40 to convert the PathMappingRulesFile and VRSceneOutputPath parameters from backslashes to forward slashes."

It converts backslashes to forward slashes — `C:\Sessions\...` becomes `C:/Sessions/....`
On Windows, both work fine for file paths. Python's os.path and open() handle both. So on a Windows farm worker it doesn't actually do anything meaningful.

`normalize_path` converts backslashes to forward slashes. On Windows workers it's a no-op since both work, but it's needed for the upcoming Linux support where backslashes in paths would cause file not found errors.

src/deadline/max_submitter/utilities/vray_executable_utils.py
ZainAallii: Correct — cross-OS handling depends on VRAY_EXECUTABLE being set on the worker. The fallback uses VRAY_FOR_3DSMAX{version}_MAIN which is Windows-only (set by V-Ray installer). Full cross-OS support (using shutil.which with platform-aware executable name) is tracked as a follow-up issue. https://github.com/aws-deadline/deadline-cloud-for-3ds-max/issues/229

src/deadline/max_submitter/utilities/vray_executable_utils.py
ZainAallii: Same answer as the previous comment.

This fallback is Windows-only — VRAY_FOR_3DSMAX{version}_MAIN is set by the V-Ray Windows installer and vray.exe is Windows-specific. For cross-OS rendering, VRAY_EXECUTABLE must be set on the worker. Full cross-OS support with platform-aware executable resolution is tracked as a follow-up issue: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/issues/229

src/deadline/max_submitter/scripts/tile_merge.py
ZainAallii: It was recommended not to use the host config and Conda together. 
The workflow which exports to vrscene on the farm will not be able to use Conda because it needs 3dsmax installation via the host configuration script. V-Ray is installed via host config script, not Conda, so we can't rely on Conda for the merge step dependencies. The pip install to a temp directory is a workaround to ensure Pillow/OpenEXR are available on the worker.

However, the workflow which uses local vrscene export can use vray and ffmpeg Conda.  

I have not made Vray Conda (it is being installed via hc script) which is why ffmpeg is not a Conda yet.

ZainAallii: Follow up issue to use Conda for step dependencies: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/issues/231

src/deadline/max_submitter/utilities/vrscene_job_submission.py
ZainAallii: "Correct, it's not used yet — movie creation is deferred to a follow-up request. https://github.com/aws-deadline/deadline-cloud-for-3ds-max/issues/230

src/deadline/max_submitter/scripts/tile_merge.py
ZainAallii: Thanks for pointing this out here's the follow up issue: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/issues/232

src/deadline/max_submitter/utilities/vrscene_job_submission.py
ZainAallii: Good catch. Currently `get_frame_range_from_string` only returns a single (start, end) tuple, so `1-10,20-30 ` would be parsed incorrectly. Our workflow uses contiguous frame ranges only — the job template uses OpenJD's range parameter which expects start-end. Non-contiguous ranges like 1-10,20-30 are not supported in this PR. Can track as a follow-up PR: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/issues/233

docs/design/vray-standalone-file-map.md
ZainAallii: You're right — V-Ray -region outputs full-size images with only the region rendered. There is a `-crop` flag ([Chaos docs](https://docs-chaos.atlassian.net/wiki/spaces/VNS/pages/115605774/V-Ray+Standalone+Command+Line+Options)) that would output actual tile-sized images, which would save storage. However, using `-crop` would require changes to the merge logic since tiles would be different sizes. Tracking as a future optimization.

ZainAallii: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/issues/234

src/deadline/max_submitter/utilities/vrscene_utils.py
ZainAallii: moved

src/deadline/max_submitter/scripts/fix_vrscene_paths.py
ZainAallii: Changed to `\\` .

seant-aws (approved):
Looks good, thanks for the followup issues. This is some good stuff !

src/deadline/max_submitter/scripts/create_movie.py
leongdl: ok, just something to think about and can be adjusted later. EG: The codec for ffmpeg and any other settings.

src/deadline/max_submitter/utilities/vrscene_utils.py
```diff
@@ -0,0 +1,179 @@
...
+import logging
+import os
+from typing import List, Tuple
+
+from pymxs import runtime as rt
+
+_logger = logging.getLogger(__name__)
+
+
+def is_vray_renderer() -> bool:
```
leongdl, line 16: nit - this is duplicated in the max_shared

leongdl (approved):
Looks ok but there are definitely som improvements to keep refactoring and improving

Issue #235, opened 2026-04-10
Feature request: Add V-Ray GPU rendering support for V-Ray Standalone workflow


ZainAallii:
### Describe the problem

The V-Ray Standalone workflow currently renders using CPU only. V-Ray Standalone 
supports GPU rendering via the `-rtEngine` command line flag, but our scripts 
(`path_mapping_render.py`, `tile_render.py`) don't expose this option. Farm workers 
with NVIDIA GPUs are underutilized.

Additionally, the `VRayExecutable` parameter is baked in at submit time as a Windows 
path (`vray.exe`), making it incompatible with Linux farm workers where the executable 
is just `vray`.

### Proposed Solution

1. Add a "Render Engine" dropdown to the V-Ray tab in the submitter UI:
   - CPU (default)
   - CUDA (all NVIDIA GPUs)
   - RTX (NVIDIA RTX GPUs only)

2. Pass the selection as a job parameter `RenderEngine` (0=CPU, 5=CUDA, 7=RTX)

3. In `path_mapping_render.py` and `tile_render.py`, add the flag when non-CPU is selected:
`-parameterOverride="SettingsOptions::rtEngine={{Param.RenderEngine}}"`

4. Resolve `VRayExecutable` at runtime on the worker instead of using the submitted 
Windows path:
```python
import os, shutil, sys
vray_cmd = (
    os.environ.get("VRAY_EXECUTABLE")
    or shutil.which("vray.exe" if sys.platform == "win32" else "vray")
)
```
This enables cross-OS rendering — export on Windows, render on Linux GPU workers.

### Reference V-Ray Standalone `-rtEngine` values [(Chaos docs)](https://docs-chaos.atlassian.net/wiki/spaces/VNS/pages/115605774):

0 — CPU (default)
5 — CUDA
7 — RTX

### Example Use Cases

1. Studio has AWS EC2 GPU instances (g4dn, p3) — enable CUDA rendering for significantly faster render times
2. Mixed fleet — Windows workers for export (3dsmaxcmd.exe), Linux GPU workers for rendering (vray)
3. Artist selects RTX engine for scenes with simple shading and heavy instancing for maximum performance

### Related
[PR #217](https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/217) (V-Ray Standalone workflow)

AlexTranAmz:
Thank you, we will look at this and prioritize accordingly

ZainAallii:
[...]

PR #244, opened 2026-04-30
feat: Add V-Ray GPU support with configurable RT engine parameters

ZainAallii:
https://github.com/aws-deadline/deadline-cloud-for-3ds-max/issues/235
V-Ray Standalone rendering on the farm only supported CPU rendering. Studios with GPU-capable farm workers (NVIDIA A10G, L4, L40S) could not take advantage of GPU acceleration for V-Ray Standalone jobs. Additionally, GPU tile rendering was producing incomplete renders (missing background/environment) because the RT engine was stopping too early due to vrscene-baked stopping conditions.

- Added a **Render Engine** dropdown to the V-Ray tab UI (CPU / CUDA / RTX) with corresponding  **Frame Timeout**, **Noise Threshold**, and **Max Samples** fields that appear when GPU engine is selected
- RT engine stopping conditions (`-rtTimeOut`, `-rtNoise`, `-rtSampleLevel`) are now passed to
  `vray.exe` when GPU engine is selected, overriding vrscene-baked limits — this ensures the
  progressive renderer runs long enough to fully render each tile including the
  background/environment
- UI fields auto-populate from the scene's V-Ray render settings at submit time
- Added `VRAY_ENGINE_CPU/CUDA/RTX` constants to `data_const.py` replacing magic numbers
- Added validation for render engine and RT parameter values
- Fixed output format bug where `create_vrscene_render_job_parameters()` was hardcoding `.png`
  regardless of user selection
- Documented GPU rendering behavior and known limitations in
  `docs/design/vray-standalone-file-map.md`

- Users can now select CUDA or RTX GPU rendering for V-Ray Standalone jobs
- GPU tile rendering now produces correct output with background/environment
- Output format is now respected for local export render jobs (was previously hardcoded to `.png`)
- RT engine parameters are configurable per-submission and persist between sessions

- Tested all combinations on AWS Deadline Cloud GPU instances (A10G, L4, L40S):
  - CUDA + 1 tile, local and farm export ✅
  - CUDA + multiple tiles, local and farm export ✅
  - RTX + 1 tile, farm export ✅
  - RTX + multiple tiles, local and farm export ✅
- Performance on heavy scene with 4 tiles (A10G): RTX ~5-6 min/tile vs CUDA ~8-9 min/tile
- 62 unit tests passing

Yes — `docs/design/vray-standalone-file-map.md` updated with GPU rendering behavior, RT
parameter explanation, and tested GPU instance table.

- [ ] Yes
- [x] No

No — additive only. CPU rendering behavior is unchanged. Default render engine is CPU (0).

----
*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute
this contribution, under the terms of your choice.*


leongdl (approved):

src/deadline/max_submitter/scripts/path_mapping_render.py
```diff
@@ -39,6 +39,10 @@ def main():
     output_dir = normalize_path(r"{{Param.OutputDir}}")
     output_filename = r"{{Param.OutputFileName}}"
     frame = r"{{Task.Param.Frame}}"
+    render_engine = int(r"{{Param.RenderEngine}}")
```
leongdl, line 42: This is backwards compatible right? Since this script is interpreted by the job bundle

leongdl (approved):

src/deadline/max_submitter/scripts/path_mapping_render.py
leongdl: Checked - it is backwards compatible since only new versions will use this embedded script.

ZainAallii: Thanks for looking into this and approving.

src/deadline/max_submitter/job_templates/vray_combined_job_template.yaml
```diff
@@ -44,3 +44,19 @@ parameterDefinitions:
 - name: RegionRows
   type: INT
   description: Number of region rows
+- name: RenderEngine
+  type: INT
+  default: 0
+  description: 'V-Ray render engine: 0=CPU, 5=CUDA, 7=RTX'
```
godobyte, line 51: Should we add `allowedValues: [0, 5, 7]` here and in the following templates? Could render fail or produce unexpected result if the value passed in is not in this list? 

If so, we should handle the case if someone submits a job bundle manually bypassing the UI and Python validation, the job won't be created with an unsupported engine value.

ZainAallii: I agree. Changes coming....

ZainAallii: changes done. Please approve.

godobyte (approved):

AlexTranAmz (approved):

leongdl (approved):

PR #255, opened 2026-06-17
feat: add Arnold renderer support tests

ZainAallii:
* test_arnold_handler.py: covers ArnoldHandler.check_renderer (sets rt.Arnold() when not active, no-op when already Arnold) and confirms inheritance of DefaultMaxHandler's action_dict.
* test_render_handler_dispatch.py: covers get_render_handler() for every renderer name including Arnold, plus the unknown-fallback path that returns DefaultMaxHandler.
* test_sanity_checks.py: explicit assertion that "Arnold" is in ALLOWED_RENDERERS.
* test_adaptor.py: parametrized schema validation across every supported renderer name (Default Scanline, ART, Corona, Redshift, Arnold, V-Ray CPU/GPU).

All 368 unit tests pass.

After Arnold renderer support landed in #254, several code paths had no unit-test coverage:

- `ArnoldHandler.check_renderer()` — the renderer-activation logic
- `get_render_handler()` — the renderer-name → handler-class dispatch (no tests existed for this function at all, even pre-Arnold)
- Presence of `"Arnold"` in the submitter's `ALLOWED_RENDERERS`
- Schema acceptance of `"Arnold"` as a valid `renderer` value

Without these tests, future refactors to the dispatch table or the schema enum could silently drop or break Arnold support. This PR adds focused regression coverage for everything #254 introduced and incidentally fills the dispatch-function gap that has existed since the project's earliest renderers were added.

Four files touched:

**`test/unit/deadline_adaptor_for_3ds_max/MaxClient/render_handlers/test_arnold_handler.py` (new)**

Mirrors `test_vray_handler.py` but simpler — `ArnoldHandler` has no env-var validation or version-selection logic. Two test methods:

- `test_arnold_handler_inherits_default_action_dict` — instantiates `ArnoldHandler` (with `pymxs.runtime.rt` patched out) and asserts the inherited `action_dict` from `DefaultMaxHandler` is intact (`start_render`, `camera`, `state_set`, etc.). Catches the case where someone overrides `__init__` and accidentally drops the dispatch table.
- `test_check_renderer_sets_arnold_when_not_active` — parametrized over four scenarios (current renderer is Default Scanline, V-Ray, Redshift, or Arnold). Asserts `check_renderer()` calls `rt.Arnold()` and assigns it to `rt.renderers.current` only when the active renderer is not already Arnold; the `already_arnold_no_op` case asserts no work is done. Mocks `rt.renderers.current` as a plain string so the handler's `str(...).split(":")[0]` comparison runs against real string semantics.

**`test/unit/deadline_adaptor_for_3ds_max/MaxClient/render_handlers/test_render_handler_dispatch.py` (new)**

Tests for `get_render_handler()` in `render_handlers/__init__.py`. Before this PR, the dispatch function had zero coverage. The file uses an `autouse` fixture that patches `pymxs.runtime.rt` across every handler module so handler constructors don't need a real 3ds Max:

- `test_get_render_handler_returns_expected_handler` — parametrized over six renderer names (`Arnold`, `Corona`, `ART_Renderer`, `Redshift_Renderer`, `Default_Scanline_Renderer`, plus an unknown name) and asserts the returned handler instance is of the correct concrete class. The unknown-renderer case asserts the fallback path returns `DefaultMaxHandler`. V-Ray is intentionally excluded because its constructor needs targeted env-var setup that's already covered in `test_vray_handler.py`.
- `test_get_render_handler_default_argument_is_scanline` — calls `get_render_handler()` with no argument and asserts the default parameter resolves to `DefaultMaxHandler`. Documents the API contract.

**`test/unit/deadline_submitter_for_3ds_max/test_sanity_checks.py` (modified)**

Added `test_arnold_is_in_allowed_renderers` to the existing `TestSanityChecks` class — a one-line `assert "Arnold" in ALLOWED_RENDERERS`. The pre-existing `test_allowed_renderers_no_substring_conflicts` already protects against name collisions across the list; Arnold is naturally distinct from every other entry, so that test continues to pass without modification.

**`test/unit/deadline_adaptor_for_3ds_max/MaxAdaptor/test_adaptor.py` (modified)**

Added `test_init_data_schema_accepts_supported_renderers` to `TestMaxAdaptor_semantic_version`. Parametrized over seven renderer names (Default Scanline, ART, Corona, Redshift, Arnold, V-Ray CPU versioned, V-Ray GPU versioned), it loads `init_data.schema.json`, sets the `renderer` field to each value in turn, and runs `jsonschema.validate(...)`. Regression coverage for the schema's `renderer.oneOf` clause — if anyone removes a renderer from the enum or breaks the V-Ray regex, this test fails.

The pre-existing `test_if_init_data_and_run_data_schema_are_changed_schema_version_is_bumped` test (which guards property names and required fields) was not modified; the Arnold enum addition in #254 didn't change the set of property names or required fields.

Tests-only PR. No runtime behaviour changes. The unit-test count grows by ~12 (four parametrized cases for the renderer-set check, six for the dispatch lookup, two singleton assertions, plus the parametrized schema test). All new tests run with the existing `hatch run test` workflow.

Going forward, any change that drops a handler from the dispatch table, removes a renderer from `ALLOWED_RENDERERS`, or tightens the init-data schema's `renderer` enum will be caught by CI.

- `hatch run test` from the repo root: 369 passed, 0 failed.
- Targeted runs against the new and modified files alone: 40 passed.
- `ruff check` and `black --check --line-length 100` on every changed file: no issues.

The `pymxs` mocks follow the existing pattern in `test_vray_handler.py` and the autouse fixture in `conftest.py`. No real 3ds Max is required to run the tests.

No documentation changes — this PR is unit tests only. The test IDs in `pytest.param(..., id="...")` make each case's intent self-describing in CI output.

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

This PR does not touch any schema files. The `init_data.schema.json` change that introduced `"Arnold"` to the `renderer` enum lives in #254.

No. Tests-only PR.

---

### DIFF
--- a/test/unit/deadline_adaptor_for_3ds_max/MaxAdaptor/test_adaptor.py
+++ b/test/unit/deadline_adaptor_for_3ds_max/MaxAdaptor/test_adaptor.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import importlib.resources
 import json
 import re
 from pathlib import Path
@@ -96,6 +97,47 @@ class TestMaxAdaptor_semantic_version:
         """Tests that the adaptor semantic version is in the expected format"""
         adaptor = MaxAdaptor(init_data)
         assert adaptor.integration_data_interface_version == SemanticVersion(major=0, minor=3)
+
+    @pytest.mark.parametrize(
+        "renderer_value",
+        [
+            "Default_Scanline_Renderer",
+            "ART_Renderer",
+            "Corona",
+            "Redshift_Renderer",
+            "Arnold",
+            "V_Ray_6_Hotfix_3",
+            "V_Ray_GPU_7",
+        ],
+        ids=[
+            "scanline",
+            "art",
+            "corona",
+            "redshift",
+            "arnold",
+            "vray_cpu_versioned",
+            "vray_gpu_versioned",
+        ],
+    )
+    def test_init_data_schema_accepts_supported_renderers(
+        self, init_data: dict, renderer_value: str
+    ) -> None:
+        """Init-data schema should accept every supported renderer name."""
+        # Load the schema as a package resource so this test is independent of
+        # repo layout (matches the production code in MaxAdaptor.adaptor, which
+        # resolves the schema dir relative to its own __file__).
+        schema_text = (
+            importlib.resources.files("deadline.max_adaptor.MaxAdaptor")
+            .joinpath("schemas/init_data.schema.json")
+            .read_text(encoding="utf-8")
+        )
+        init_data_schema = json.loads(schema_text)
+
+        payload = dict(init_data)
+        payload["renderer"] = renderer_value
+
+        # Should not raise
+        jsonschema.validate(payload, init_data_schema)
 
     def test_if_init_data_and_run_data_schema_are_changed_schema_version_is_bumped(
         self, init_data: dict
--- a/test/unit/deadline_adaptor_for_3ds_max/MaxClient/render_handlers/test_arnold_handler.py
+++ b/test/unit/deadline_adaptor_for_3ds_max/MaxClient/render_handlers/test_arnold_handler.py
@@ -1,0 +1,47 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+from __future__ import annotations
+
+from unittest.mock import patch
+
+import pytest
+
+
+class TestArnoldHandler:
+    """Tests for ArnoldHandler renderer activation."""
+
+    @pytest.mark.parametrize(
+        "current_renderer_str,expected_set_called",
+        [
+            pytest.param(
+                "Default_Scanline_Renderer:Default_Scanline_Renderer", True, id="from_scanline"
+            ),
+            pytest.param("V_Ray_6_Hotfix_3:V_Ray_6_Hotfix_3", True, id="from_vray"),
+            pytest.param("Redshift_Renderer:Redshift_Renderer", True, id="from_redshift"),
+            pytest.param("Arnold:Arnold", False, id="already_arnold_no_op"),
+        ],
+    )
+    def test_check_renderer_sets_arnold_when_not_active(
+        self, current_renderer_str: str, expected_set_called: bool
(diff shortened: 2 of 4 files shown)


test/unit/deadline_adaptor_for_3ds_max/MaxClient/render_handlers/test_arnold_handler.py
```diff
@@ -0,0 +1,71 @@
...
+
+from unittest.mock import patch
+
+import pytest
+
+
+class TestArnoldHandler:
+    """Tests for ArnoldHandler renderer activation."""
+
+    def test_arnold_handler_inherits_default_action_dict(self) -> None:
```
jericht: Do we do this for the other tests? It may be a pain to update many test files each time we add a new action to the default handler

test/unit/deadline_adaptor_for_3ds_max/MaxAdaptor/test_adaptor.py
```diff
@@ -97,6 +97,44 @@ def test_semantic_version(self, init_data: dict) -> None:
...
+        ],
+    )
+    def test_init_data_schema_accepts_supported_renderers(
+        self, init_data: dict, renderer_value: str
+    ) -> None:
+        """Init-data schema should accept every supported renderer name."""
+        root_directory_path = Path(__file__).parent.parent.parent.parent.parent
+        schema_path = root_directory_path.joinpath(
+            "src", "deadline", "max_adaptor", "MaxAdaptor", "schemas", "init_data.schema.json"
+        )
```
jericht: This is prone to failing if the files in this repo are shifted around.

test/unit/deadline_adaptor_for_3ds_max/MaxClient/render_handlers/test_arnold_handler.py
ZainAallii: Fixed.

leongdl (approved):

jericht (approved):

PR #261, opened 2026-07-07
fix(installer): Stop polluting global PYTHONPATH and clean up prior installs

### LINKED ISSUE #225: Bug: Installer PYTHONPATH interferes with global deadline install
### Describe the bug

When I use the global installer to install the 3DSMax submitter, it sets the PYTHONPATH environment variable to `C:\Program Files\DeadlineCloudFor3dsMaxSubmitter/scripts`, and that seems to be interfering with the system-wide install of the deadline client. 

### Expected Behaviour

Running `deadline bundle gui-submit` from the command line allows me to submit a job. 

### Current Behaviour

Running `deadline bundle gui-submit` from the command line throws this error:
```
File "C:\Program Files\Python313\Lib\concurrent\futures\_base.py", line 401, in __get_result
    raise self._exception
  File "C:\Program Files\Python313\Lib\concurrent\futures\thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "D:\Users\olly\AppData\Roaming\Python\Python313\site-packages\deadline\job_attachments\upload.py", line 1101, in _process_input_path
    entry.file_hash = hash_file(full_path, hash_alg)
                      ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "D:\Users\olly\AppData\Roaming\Python\Python313\site-packages\deadline\job_attachments\asset_manifests\hash_algorithms.py", line 27, in hash_file
    from xxhash import xxh3_128
  File "C:\Program Files\DeadlineCloudFor3dsMaxSubmitter\scripts\xxhash\__init__.py", line 1, in <module>
    from ._xxhash import (
    ...<17 lines>...
    )
ModuleNotFoundError: No module named 'xxhash._xxhash'
```
It seems to be trying to load xxhash from the 3DSMax submitter install. 

Clearing the PYTHONPATH environment variable allows the `deadline bundle submit` command to function normally.

### Reproduction Steps

1. Install deadline with `pip install deadline`
2. Run a system-wide installation of the deadline for 3DSMax submitter
3. Run `deadline bundle submit .` on a job bundle directory

### Environment

At minimum:
1. Operating system Windows 11
2. Version of 3ds Max: 2024
3. Version of this package: 0.2.0

Please share other details about your environment that you think might be relevant to reproducing the bug.

leongdl:
Fixes: https://github.com/aws-deadline/deadline-cloud-for-3ds-max/issues/225

The installer wrote `${max_installdir}/scripts` to the **global** `PYTHONPATH` environment variable (user or system scope, persisted to the Windows registry). That directory contains the submitter's bundled `deadline` package, so it leaked into every Python interpreter on the machine

**1. Stop writing global `PYTHONPATH` in the first place** (`installer/deadline-cloud-for-3ds-max.xml`).
The write was redundant. 3ds Max already runs `DeadlineCloudMenu.ms` at startup (via `ADSK_3DSMAX_STARTUPSCRIPTS_ADDON_DIR`), and that script prepends the submitter's dirs to `sys.path` using the `MAX_SUBMITTER` env var. `MAX_SUBMITTER` is 3ds-Max-specific and harmless because Python doesn't read it.

**2. Clean up already-polluted machines on upgrade** (`installer/deadline-cloud-for-3ds-max.xml`, `postInstallationActionList`).
A new `fnRemovePathFromEnvironmentVariable` action is invoked for both user and system scopes at the start of `postInstall`, so customers who ran an earlier installer just need to run the new one — they do NOT need to uninstall 3ds Max first.

**3. Clean up on uninstall too** (`installer/DeadlineCloudFor3dsMaxSubmitter.xml`, `postUninstallationActionList`).
Same action is called during uninstall so removing the submitter leaves the machine clean.

**4. New generic action fnRemovePathFromEnvironmentVariable** (installer/DeadlineCloudFor3dsMaxSubmitter.xml). Windows-only implementation via PowerShell. Comparison is case-insensitive and slash-insensitive so entries stored with either forward or back slashes are matched. Values are passed as PowerShell script parameters (typed param() block), not string-interpolated into the command text. Best-effort semantics: PowerShell body is wrapped in try/catch, and the enclosing <runProgram> sets abortOnError=0 and showMessageOnError=0. If anything unusual happens (e.g., a user-scope install attempting a system-scope registry write and getting a SecurityException, or PowerShell not being available on some minimal Windows SKU), cleanup is silently skipped — the install/uninstall itself never fails or shows an error dialog because of this action.

**5. Defense in depth: prepend to `sys.path` instead of append** (`install_files/DeadlineCloudMenu.ms`).
`sys.path[0:0] = list(...)` with dedup, so the bundled `deadline` package reliably wins even if the user's global Python has an older `deadline` installed via pip.

**6. Defense in depth: bootstrap `sys.path` in `run_ui.py`** (`src/deadline/max_submitter/run_ui.py`).
Small idempotent block at module load that prepends the submitter root to `sys.path` if it isn't already there, so the submitter opens even in the corner case where `DeadlineCloudMenu.ms` didn't run (customer disabled startup scripts, etc.).

- **For customers with only 3ds Max**: transparent — the submitter still works exactly the same. `sys.path` inside 3ds Max is still populated the same way; only the global env-var side-effect is gone.
- **For customers with 3ds Max + Maya (or Nuke, or Blender)**: fixes the path pollution issue that could cause issue when multiple submitters are installed on the same machine.
- **For customers who already installed a broken version**: running this installer over their existing install cleans up the polluted `PYTHONPATH` automatically. No 3ds Max uninstall required. Uninstalling the submitter now also cleans it.
- **Persistence semantics**: `max_installdir` is now persisted (`persist="1"`) so the uninstaller can reference it — that's a minor behavior refinement, no customer-visible impact.

- XML well-formed check passes for both installer XMLs.
- `black --check` and `ruff check` clean on `src/deadline/max_submitter/run_ui.py`.
- **`hatch run test` — all 368 unit tests pass with the change applied** (unit test suite covers the adaptor and submitter Python code paths; there are no unit tests for the InstallBuilder XML actions, per the existing repo convention).
- Executed the exact Python sequence emitted by `DeadlineCloudMenu.ms` in a real interpreter and confirmed `sys.path` ordering matches the previous behavior (submitter root first, then `.../deadline/max_submitter`) and that entries dedup on re-run.
- Existing integration tests reviewed: they inject `PYTHONPATH` themselves as **process-scoped** test scaffolding (`os.environ["PYTHONPATH"] = ...` before spawning 3ds Max). That's independent from the persistent user/system-scope write we removed, so nothing in the test suite depends on the behavior being removed.

No user-facing docs are added — the change is invisible in normal use. Suggest calling out the cleanup behavior briefly in the next release notes so support-facing folks know upgrade-in-place is the recommended fix for existing broken installs. The release process manages `CHANGELOG.md` so no edit there in this PR.

- [ ] Yes
- [x] No

   - [ ] Yes
   - [ ] No

No. Behavior visible to 3ds Max and to the submitter UI is unchanged. The only removed behavior is the accidental side-effect on global `PYTHONPATH`, which was never a documented or intentional feature.

----

### DIFF
--- a/install_files/DeadlineCloudMenu.ms
+++ b/install_files/DeadlineCloudMenu.ms
@@ -4,12 +4,25 @@
 global maxVersion_ = maxVersion()
 global majorVersion = maxVersion_[1]
 
--- Python script to add repo path to max script locations
-python.Execute "import sys"
-python.Execute "import os"
-python.Execute "max_submitter_path: str = os.environ.get('MAX_SUBMITTER', '')"
-python.Execute "sys.path.append(max_submitter_path)"
-python.Execute "sys.path.append(f'{max_submitter_path}/deadline/max_submitter')"
+-- Python script to add the submitter's own directories to sys.path.
+-- Prepend (insert at 0) rather than append so the bundled 'deadline' package always wins
+-- over any leaked older 'deadline' that may be present in a user or system Python install.
+-- Use os.path.normpath / os.path.join so the entries are stored in canonical native form
+-- (matching what run_ui.py's defense-in-depth bootstrap uses), keeping the exact-string
+-- dedup below correct on Windows regardless of how the MAX_SUBMITTER env var was written
+-- by the installer.
+--
+-- IMPORTANT: read the raw env value and truthy-check it BEFORE calling os.path.normpath.
+-- os.path.normpath("") returns "." (truthy), which would otherwise cause "." and
+-- "./deadline/max_submitter" to be prepended to sys.path if MAX_SUBMITTER were unset —
+-- exactly the CWD-on-sys.path footgun this whole change is trying to eliminate.
+python.Execute "import sys, os"
+python.Execute "_dc_max_submitter_raw = os.environ.get('MAX_SUBMITTER', '')"
+python.Execute "_dc_max_submitter_path = os.path.normpath(_dc_max_submitter_raw) if _dc_max_submitter_raw else ''"
+python.Execute "_dc_sys_path_entries = tuple((_dc_max_submitter_path, os.path.join(_dc_max_submitter_path, 'deadline', 'max_submitter'))) if _dc_max_submitter_path else ()"
+python.Execute "sys.path[:] = [p for p in sys.path if p not in _dc_sys_path_entries]"
+python.Execute "sys.path[0:0] = list(_dc_sys_path_entries)"
+python.Execute "del _dc_max_submitter_raw, _dc_max_submitter_path, _dc_sys_path_entries"
 
 -- Check if this is 3ds Max 2024 or earlier (version 26000 and below)
 if majorVersion <= 26000 then
--- a/installer/DeadlineCloudFor3dsMaxSubmitter.xml
+++ b/installer/DeadlineCloudFor3dsMaxSubmitter.xml
@@ -83,33 +83,13 @@
         <actionDefinition name="fnRemovePathFromEnvironmentVariable">
             <!--
                 Removes a single directory entry from a PATH-like environment variable
-                (semicolon-separated on Windows). Windows-only implementation via PowerShell;
-                no-op on other platforms (3ds Max is Windows-only).
+                using InstallBuilder-native actions only. Avoid runProgram/PowerShell
+                because those are brittle on Windows and can be blocked or skipped.
 
-                The primary user of this action is one-time cleanup of a bug introduced in
-                earlier releases of the 3ds Max submitter installer, which added the submitter
-                scripts directory to the GLOBAL PYTHONPATH env var. That leaked into every Python
-                process on the machine (including Maya, Nuke, etc.) and caused the bundled
-                'deadline' package from 3ds Max to shadow the correct one shipped with those
-                other DCC submitters.
-
-                Parameter marshalling: user-controlled values (`name`, `value`, `scope`) are
-                passed to the PowerShell child by writing them to the installer's own process
-                environment via <setEnvironmentVariable> — a non-persistent action that lives
-                only for the current installer run and is inherited by children launched via
-                <runProgram>. The PowerShell command text itself contains no interpolation of
-                user-controlled data, so install paths containing shell-meaningful characters
-                (single quotes, `$`, etc.) cannot break parsing.
-
-                Comparison inside PowerShell is case-insensitive and slash-insensitive to match
-                Windows path semantics, so entries stored with either forward or back slashes
-                are removed.
-
-                Best-effort: PowerShell body is wrapped in try/catch, and the runProgram sets
-                abortOnError=0 / showMessageOnError=0. If anything unusual happens (e.g., a
-                user-scope install attempting a system-scope registry write and getting a
-                SecurityException), the cleanup is silently skipped rather than failing the
-                enclosing install or uninstall.
+                The regular expression removes the target when it appears as the
+                only entry, first entry, middle entry, or last entry in a semicolon
+                separated value. If the result is empty, delete the environment
+                variable; otherwise write back the cleaned value.
             -->
             <parameterList>
                 <stringParameter name="name"/>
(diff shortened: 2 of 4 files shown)


installer/DeadlineCloudFor3dsMaxSubmitter.xml
```diff
@@ -80,6 +80,72 @@
...
+                            <text>${env(${name})}</text>
+                            <pattern>(^|;)${_regex_safe_value}(;|$)</pattern>
+                            <substitution>\1</substitution>
+                        </setInstallerVariableFromRegEx>
+                        <setInstallerVariableFromRegEx>
+                            <name>_cleaned_env_value</name>
+                            <text>${_cleaned_env_value}</text>
+                            <pattern>;$</pattern>
+                            <substitution></substitution>
+                        </setInstallerVariableFromRegEx>
```
jericht, line 143: Neat, seems to look right. Handles beginning, middle, and end of PATH env vars it looks like

jericht (commented):
Before we merge, I think we should test on a real installation if we haven't already.

installer/DeadlineCloudFor3dsMaxSubmitter.xml
leongdl: :) Yeah! Thats what I was struggling with all afternoon

install_files/DeadlineCloudMenu.ms
```diff
@@ -4,12 +4,25 @@
 global maxVersion_ = maxVersion()
```
leongdl, line 4: Here's the testing that I did:

## Testing Summary

**Test 1 — Single entry (only stale path present)**
- Set `PYTHONPATH` (User) = `C:\Users\RDP\DeadlineCloudFor3dsMaxSubmitter/scripts`
- Ran installer → PYTHONPATH deleted entirely ✅

**Test 2 — Multi-entry (stale path in the middle)**
- Set `PYTHONPATH` (User) = `C:\KeepMe;C:\Users\RDP\DeadlineCloudFor3dsMaxSubmitter/scripts;C:\AlsoKeep`
- Ran installer → PYTHONPATH = `C:\KeepMe;C:\AlsoKeep` ✅

**Test 3 — Path with parentheses (regex metacharacter case from PR feedback)**
- Set `PYTHONPATH` (User) = `C:\KeepMe;C:\Users\RDP\DeadlineCloudFor3dsMaxSubmitter (x86)/scripts;C:\AlsoKeep`
- Ran installer (targeting same `(x86)` path) → PYTHONPATH = `C:\KeepMe;C:\AlsoKeep` ✅

**What was confirmed:**
- Native InstallBuilder cleanup (`setInstallerVariableFromRegEx` + `deleteEnvironmentVariable` / `addEnvironmentVariable`) works reliably — no PowerShell, no `runProgram`
- Regex metacharacters in paths (parentheses, dots, etc.) are properly escaped before pattern matching
- Unrelated PYTHONPATH entries are preserved intact
- Cleanup fires on both install (upgrade-in-place) and uninstall

leongdl:
> Before we merge, I think we should test on a real installation if we haven't already.

## Testing Summary

**Test 1 — Single entry (only stale path present)**
- Set `PYTHONPATH` (User) = `C:\Users\RDP\DeadlineCloudFor3dsMaxSubmitter/scripts`
- Ran installer → PYTHONPATH deleted entirely ✅

**Test 2 — Multi-entry (stale path in the middle)**
- Set `PYTHONPATH` (User) = `C:\KeepMe;C:\Users\RDP\DeadlineCloudFor3dsMaxSubmitter/scripts;C:\AlsoKeep`
- Ran installer → PYTHONPATH = `C:\KeepMe;C:\AlsoKeep` ✅

**Test 3 — Path with parentheses (regex metacharacter case from PR feedback)**
- Set `PYTHONPATH` (User) = `C:\KeepMe;C:\Users\RDP\DeadlineCloudFor3dsMaxSubmitter (x86)/scripts;C:\AlsoKeep`
- Ran installer (targeting same `(x86)` path) → PYTHONPATH = `C:\KeepMe;C:\AlsoKeep` ✅

**What was confirmed:**
- Native InstallBuilder cleanup (`setInstallerVariableFromRegEx` + `deleteEnvironmentVariable` / `addEnvironmentVariable`) works reliably — no PowerShell, no `runProgram`
- Regex metacharacters in paths (parentheses, dots, etc.) are properly escaped before pattern matching
- Unrelated PYTHONPATH entries are preserved intact
- Cleanup fires on both install (upgrade-in-place) and uninstall

I will test as soon as we do a build as well.

jericht (dismissed):
Confirmed testing

installer/DeadlineCloudFor3dsMaxSubmitter.xml
```diff
@@ -189,5 +255,24 @@
         </actionGroup>
         <deleteFile path="${installdir}/installer_version.txt"/>
     </preUninstallationActionList>
+    <postUninstallationActionList>
+        <!--
+            After uninstall, sweep any stale PYTHONPATH entry that a pre-fix installer may have
+            left in the user's or system's environment. Safe no-op if not present.
```
epmog, line 282: safe no-op if the user doesn't have permissions to modify the system variable? (not running as Admin)

leongdl: Good question, I'll double check this.

leongdl: Thanks for the reviews. I tested all three concerns on a live Windows workstation (real install/uninstall, values set via [Environment]::SetEnvironmentVariable, read back in a fresh shell). Summary of each question and what the testing showed:

@jericht — "test on a real installation before we merge"
Done. Verified end-to-end on a real Windows install/uninstall, not just a build. Covered single-entry, multi-entry, and paths containing regex metacharacters:

Single entry (PYTHONPATH = stale only) → variable removed entirely.
Middle entry (KeepMe;stale;AlsoKeep) → only the stale entry stripped, KeepMe;AlsoKeep preserved, ; seams intact.
Path with parentheses (C:\…\Submitter (x86)\scripts) → removed correctly. This confirms the regex-metacharacter escaping works (without it, (x86) would be treated as a capture group and cleanup would silently no-op).
@epmog (Morgan) — "safe no-op if the user isn't admin and can't modify the system variable?"
Yes, confirmed safe. I forced the system-scope cleanup to run under a restricted (non-admin) token with a polluted machine PYTHONPATH:

Install completed with no error dialog and no abort.
The machine (HKLM) variable was not corrupted.
The env write actions are now abortOnError=0 / showMessageOnError=0, so a system-scope (HKLM) write the user lacks rights for is a silent no-op rather than a failed install. Net: a non-admin user install simply leaves machine-scope pollution in place (which only an admin could remove anyway) without breaking the install.

github-actions (Claude review) — "reads ${env(PYTHONPATH)} (merged user+system) but writes to a specific scope → can clobber/inject across scopes"
This was a real, reproducible bug, and it's now fixed.

Root cause confirmed: the old installer added the scripts dir with <scope>${installscope}</scope>, so all-users installs polluted machine-scope PYTHONPATH. On such a machine, reading the merged env() and writing per-scope caused cross-scope bleed.

Reproduction (stale entry in machine scope, user scope empty):

Before the fix: the user-scope branch read the merged env (which reflected the machine value), stripped the stale entry, and wrote the leftover into user scope — injecting a machine value into the user PYTHONPATH — while the actual machine pollution was left untouched. Observed: USER=[C:\LegitKeep], MACHINE unchanged. ❌
After the fix: each scope's value is read from its own registry key (HKCU\Environment for user, HKLM\…\Session Manager\Environment for system) via registryGet, and all matching/removal operates on that scope-specific value. Observed on an admin install: USER=[], MACHINE=[C:\LegitKeep]. ✅
I verified the mechanism with an install-time diagnostic: the user-scope read returned [] from HKCU (so nothing to inject), and the system-scope read returned the polluted value from HKLM (so it was correctly cleaned).

What changed in the fix
Native InstallBuilder actions only (no runProgram/PowerShell).
Regex metacharacters in the path are escaped before matching.
Values are read from the scope-specific registry key instead of the merged process environment — eliminates cross-scope bleed.
Env writes are non-fatal so a non-admin install isn't aborted by an HKLM write.

epmog (approved):

jericht (approved):

PR #266, opened 2026-07-09
feat: add Arnold AOV support

ZainAallii:
Wire Arnold AOVs through the render handler so per-frame AOV filenames are unique and AOV_Manager.outputPath is remapped to the worker's output directory. Arnold AOVs are managed by renderer.AOV_Manager (not the standard render element manager), so the handler drives the manager directly.

Arnold AOVs did not render correctly for animations through the adaptor, for two reasons:

1. **Filename collisions.** Arnold writes one file per `AOV_Manager` driver, and each file's basename comes from that driver's `filenameSuffix` (e.g. `AOVs.exr`). There is no per-frame token in the name, so every frame overwrote the previous frame's AOV output — a hard blocker for animation.
2. **Hard-coded output path.** `AOV_Manager.outputPath` is baked into the scene at author time and points at the artist's local filesystem, which isn't writable on a Deadline Cloud worker.

Unlike V-Ray render elements, Arnold AOVs are managed by `renderer.AOV_Manager` (an `ArnoldAOVsManager` reference target), not by 3ds Max's standard `RenderElementManager`, so this required driving the AOV manager directly from the handler.

- `_apply_path_mapping` remaps `AOV_Manager.outputPath` through the adaptor's `map_path` after scene load.
- `_configure_renderer_output` sets `AOV_Manager.outputPath` to the job's output directory and rewrites each driver's `filenameSuffix` to `<output_name>_<original_suffix>`, giving every frame a unique per-driver filename. If a driver's suffix is empty, `_AOV` is used as a safety tag so the AOV file doesn't collide with the beauty pass.
- Original suffixes are cached per driver index and restored via `_restore_arnold_aov_drivers` (invoked from the existing `cleanup_render_elements` framework hook). A `try/finally` guarantees the cache is cleared even if a restore raises. The cache also prevents suffix compounding across frames within a session.
- `check_renderer` uses `startswith("Arnold")` for version robustness.
- Every defensive `except Exception` logs a warning; no silent failures.
- Beauty pass rendering is unchanged — `_configure_renderer_output` returns `False`, so the framework's existing `rt.render(...)` call still writes the main output. (Contrast with `VrayHandler`, which returns `True` in raw-output modes because V-Ray writes the beauty pass into its own container.)

- Multi-frame Arnold jobs with AOVs now produce one file per frame per driver, instead of each frame overwriting the last.
- Arnold AOV output lands in the job's output directory on workers, where job attachments can pick it up.
- No change to the beauty pass path, and no change to jobs that don't use AOVs.

- 16 new unit tests in `test_arnold_aov.py` covering path remapping (including `"undefined"` and unchanged paths), per-frame suffix mutation, the empty-suffix `_AOV` fallback, once-only caching / no compounding across frames, cleanup restoration, and the always-clear-cache path. Plus a versioned `check_renderer` test case in `test_arnold_handler.py`.
- `hatch run test`: full suite passes.
- OpenJD runner locally: 3 frames × a 3-driver scene produced frame-unique files with no collisions.
- Deadline Cloud submission: rendered `globe_rotation.max` (2 AOV drivers) across multiple frames. Per-frame logs confirm `outputPath` remapping and per-driver `filenameSuffix` mutation with no compounding; all frames exited 0 and uploaded to S3.

Inline docstrings on the class (explaining the divergence from V-Ray's render-element path) and on each helper method. No user-facing docs required.

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

No. The beauty pass is unchanged and non-AOV jobs are unaffected. Existing AOV jobs previously produced only the last frame's output; they now correctly produce one file per frame per driver.

----

### DIFF
--- a/src/deadline/max_adaptor/MaxClient/render_handlers/arnold_handler.py
+++ b/src/deadline/max_adaptor/MaxClient/render_handlers/arnold_handler.py
@@ -1,10 +1,11 @@
 """
-3ds Max Deadline Cloud Adaptor - Arnold specific actions
+3ds Max Deadline Cloud Adaptor - Arnold (MAXtoA) specific actions
 
 Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 """
 
 import sys
+from typing import Any, List, Optional
 
 import pymxs  # noqa
 from pymxs import runtime as rt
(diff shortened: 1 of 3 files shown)


src/deadline/max_adaptor/MaxClient/render_handlers/arnold_handler.py
```diff
@@ -17,18 +18,255 @@
...
+        if mapped == original:
+            return
+
+        try:
+            aov_mgr.outputPath = mapped
+            self.log_to_console(f"Remapped Arnold AOV outputPath: {original} -> {mapped}")
+        except Exception as exc:
+            self.log_to_console(f"Warning: failed to remap Arnold AOV outputPath: {exc}")
+
+    def _configure_renderer_output(
```
jericht, line 97: This function always returns False, is that intentional? Docstring mentions it is supposed to return true in raw output modes

src/deadline/max_adaptor/MaxClient/render_handlers/arnold_handler.py
```diff
@@ -17,18 +18,255 @@
...
+        we can't give a more precise annotation without a stub. Callers should
+        treat ``None`` as "AOV manager not accessible" and no-op.
+        """
+        try:
+            return rt.renderers.current.AOV_Manager
+        except Exception as exc:
+            self.log_to_console(f"Warning: could not access Arnold AOV_Manager: {exc}")
+            return None
+
+    def _iter_drivers(self, aov_mgr: Any) -> Iterator[Tuple[int, Any]]:
```
jericht: This function seems unnecessary. Can callers not directly iterate over `aov_mgr.drivers` themselves? It looks like it is already a list

src/deadline/max_adaptor/MaxClient/render_handlers/arnold_handler.py
```diff
@@ -17,18 +18,255 @@
...
+        # first time _configure_renderer_output runs. Read on every frame to
+        # compose the new suffix from the artist's original value (avoids
+        # compounding suffixes across frames within a session) and drained by
+        # _restore_arnold_aov_drivers at session end.
+        #
+        # NOTE: This assumes the driver list does not change during a session.
+        # A single render job's drivers array is expected to be static; if a
+        # scenario emerges where drivers are added or removed mid-session,
+        # switch to keying by driver object identity.
+        self._arnold_aov_driver_original_suffixes: Dict[int, str] = {}
```
jericht: nit: we can simplify this to a `list` and just copy the original suffixes to it, restoring it in the same order after we're done

leongdl (changes requested):
Thanks for this contribution — it fixes two real blockers for Arnold animation jobs (per-frame AOV filename collisions and the scene-baked `outputPath`), and the quality bar here is high: the docstrings explaining why Arnold diverges from the V-Ray render-element path are exactly what a future maintainer needs, `_iter_drivers` cleanly quarantines the defensive pymxs access pattern, the cache-once-across-frames test is the right regression test, and the verification story (OpenJD runner + real Deadline Cloud submission) is stronger than most PRs. 

I traced the three flows (scene load → `_apply_path_mapping`; per-frame `start_render` → `_configure_renderer_output`; session end → `cleanup_render_elements` → restore) and have three changes to request plus one recommendation.

### Requested changes

**1. Asymmetric pymxs `undefined` handling can leak the literal string `"undefined"` into filenames and the restore cache**

`_apply_path_mapping` explicitly normalizes pymxs's `"undefined"` stringification, but the suffix read in `_configure_renderer_output` uses `str(drv.filenameSuffix or "")`. pymxs `undefined` is not guaranteed falsy across MAXtoA versions — if it's truthy, a driver with an unset suffix produces files named `frame0001_undefined.exr`, and worse, `"undefined"` gets cached and **written back into the scene** at cleanup. Suggest a small shared helper (e.g. `_normalize_mxs_str(value) -> str` that returns `""` for `None`/`"undefined"`) used in both places.

**2. Two empty-suffix drivers still collide with each other**

The `_AOV` safety tag correctly distinguishes empty-suffix AOV output from the beauty pass, but two drivers that *both* have empty suffixes compose to the same `<output_name>_AOV.<ext>` and overwrite each other — the exact failure mode this PR exists to fix, just one level down. Suggest including the driver index in the fallback: `f"{output_name}_AOV{i}"`.

**3. Misleading log: the `_apply_path_mapping` remap is always clobbered before any render**

Every render path goes through `start_render` → `_configure_renderer_output`, which unconditionally sets `aov_mgr.outputPath = output_dir` (and `start_render` raises earlier if `output_dir` is unset). So the remapped value never survives to a render — yet the code logs `"Remapped Arnold AOV outputPath: X -> Y"`, which will send anyone debugging output paths from worker logs chasing a value that was overwritten. Either drop the outputPath remap from `_apply_path_mapping`, or keep it as an explicit documented fallback and reword the log so it can't be mistaken for the effective render-time path (e.g. `"...(will be overridden by job output directory at render time)"`).

### Recommended (non-blocking)

- The PR description mentions tests for the `"undefined"` and unchanged-path remap cases, but I don't see them in the diff — adding them would also pin the fix for item 1.
- Test helper `_make_handler_with_aov_manager` starts the `rt` patcher before returning it; if anything raises between `start()` and the caller's `try` (e.g. handler construction), `rt` stays patched for the rest of the test session. A pytest fixture or `@contextmanager` yielding `(handler, aov_mgr, drivers)` fixes that and deletes all eight `try/finally patcher.stop()` blocks.

Edge cases I checked and cleared, so you don't need to re-investigate: non-Arnold renderer active at map time (caught, warns, no-ops), multi-frame suffix compounding (prevented by cache-once, tested), restore raising mid-loop (`finally` clears the cache), and `output_dir` creation ordering (base creates it after the hook, before `rt.render`).

Happy to approve once items 1–3 are addressed — they're all small, same-file changes.

🤖 *AI-assisted review* 🤖

leongdl:
Summary. (me trying to understand the problem)
  
  PR #266 fixes two real blockers for Arnold animation jobs: per-frame AOV filename collisions (every frame overwrote the last, since Arnold names AOV files purely from
  each driver's filenameSuffix) and the scene-baked AOV_Manager.outputPath pointing at the artist's local disk. Because Arnold AOVs live in renderer.AOV_Manager rather than 3ds Max's RenderElementManager, the
  handler drives the manager directly instead of reusing the V-Ray render-element path — that architectural choice is correct and well documented.
  
  Call-stack traces
  
  Flow 1 — scene load / path mapping:
  
  1. MaxClient action "scene_file"
  2. DefaultMaxHandler.set_scene_file          rt.loadMaxFile(...)
  3.   if self.map_path is not None:           (base guards None already)
  4. ArnoldHandler._apply_path_mapping         re-checks map_path (redundant, harmless)
  5.   _get_aov_manager → rt.renderers.current.AOV_Manager
  6.   read outputPath → map_path() → write back
  
  Flow 2 — per frame:
  
  1. action "start_render" {frame}
  2. DefaultMaxHandler.start_render            resolves tokens, frame padding, basename
  3. ArnoldHandler._configure_renderer_output  sets AOV outputPath = output_dir;
                                               per driver: cache original suffix ONCE,
                                               set suffix = f"{output_name}_{orig}"
  4. returns False → base makes output_dir, calls rt.render(camera, outputFile=...)
  
  Flow 3 — session end:
  
  1. action "cleanup_render_elements"
  2. ArnoldHandler.cleanup_render_elements
  3.   _restore_arnold_aov_drivers             restore suffixes, finally: cache.clear()
  4.   super().cleanup_render_elements         standard render-element cleanup

src/deadline/max_adaptor/MaxClient/render_handlers/arnold_handler.py
ZainAallii: Intentional — Arnold's handler only redirects the AOV drivers (sets AOV_Manager.outputPath and per-frame filenameSuffix); it never takes over the beauty pass. The beauty pass is still written by the framework via rt.render(camera=..., outputFile=...), which only happens when this returns False. Returning True would suppress the beauty output entirely.

The "returns True in raw-output modes" line in the docstring is describing VrayHandler as a contrast (V-Ray writes the beauty pass itself in raw/split-buffer modes, so it must suppress the framework's outputFile to avoid duplicates), not this method. I'll reword the docstring so that's unambiguous.

src/deadline/max_adaptor/MaxClient/render_handlers/arnold_handler.py
```diff
@@ -17,18 +18,253 @@
...
+                        f"Warning: could not read Arnold driver[{i}].filenameSuffix: {exc}"
+                    )
+                    self._arnold_aov_driver_original_suffixes.append("")
+
+            original_suffix = self._arnold_aov_driver_original_suffixes[i]
+            # Fall back to "_AOV<index>" when the driver's suffix is empty so
+            # its files collide neither with the beauty pass nor with another
+            # empty-suffix driver.
+            new_suffix = (
+                f"{output_name}_{original_suffix}" if original_suffix else f"{output_name}_AOV{i}"
```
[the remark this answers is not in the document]
ZainAallii: fixed

src/deadline/max_adaptor/MaxClient/render_handlers/arnold_handler.py
```diff
@@ -17,18 +18,206 @@
...
+        main render output. (``VrayHandler``, by contrast, returns ``True`` in
+        raw-output modes because V-Ray writes the beauty pass into its own
+        multichannel container file and the framework must not write it again.)
+        """
+        aov_mgr = self._get_aov_manager()
+        if aov_mgr is None:
+            return False
+
+        try:
+            aov_mgr.outputPath = output_dir
```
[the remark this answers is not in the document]
ZainAallii: Fixed.

src/deadline/max_adaptor/MaxClient/render_handlers/arnold_handler.py
```diff
@@ -17,18 +18,247 @@
...
+        try:
+            if aov_mgr is not None:
+                if self._arnold_aov_original_output_path is not None:
+                    try:
+                        aov_mgr.outputPath = self._arnold_aov_original_output_path
+                    except Exception as exc:
+                        self.log_to_console(
+                            f"Warning: failed to restore Arnold AOV_Manager.outputPath: {exc}"
+                        )
+                for i, drv in enumerate(aov_mgr.drivers):
```
AlexTranAmz: `aov_mgr.drivers` is the one pymxs access in this method not wrapped in an `except`. If it's throws here, the `finally` clears the cache but the exception escapes before `super().cleanup_render_elements(data)` runs — so the standard render-element restore silently gets skipped

src/deadline/max_adaptor/MaxClient/render_handlers/arnold_handler.py
```diff
@@ -17,18 +18,247 @@
...
+        Session-end framework hook.
+
+        The base class's cleanup handles standard 3ds Max render elements.
+        Arnold AOVs are a separate subsystem (``AOV_Manager``, not the render
+        element manager), but the adaptor only exposes this single per-session
+        cleanup hook — so both live here. The AOV work is delegated to
+        ``_restore_arnold_aov_drivers`` so its name reflects what it actually
+        does.
+        """
+        self._restore_arnold_aov_drivers()
```
AlexTranAmz: this cleanup hook only gets enqueued when `enabled_modify_render_elements` is true (see `on_cleanup` in `adaptor.py`). So for a typical Arnold AOV job, this restore never actually runs, is this what we expected?

src/deadline/max_adaptor/MaxClient/render_handlers/arnold_handler.py
```diff
@@ -1,10 +1,11 @@
 """
```
AlexTranAmz, line 1: The description and the call-stack comment above both mention an `_apply_path_mapping override` that isn't actually in the diff anymore? 

None of the test touch path remapping (the "undefined" one is about suffix normalization). Mind adding some?

leongdl:
Re: the cleanup-hook gating question ([discussion](https://github.com/aws-deadline/deadline-cloud-for-3ds-max/pull/266#discussion_r3686674999)) — I traced this through `adaptor.py:on_cleanup` and confirmed the behavior: the `cleanup_render_elements` action is only enqueued when `enabled_modify_render_elements == "true"`, so for a typical Arnold AOV job the `_restore_arnold_aov_drivers` path never runs.

That said, I consider this **non-blocking**. The gated cleanup means the restore machinery is *dead*, not *wrong*: on a worker the scene copy is discarded at session end anyway, so the missing restore has zero production impact today. Its only cost is misleading docstrings that promise a restore guarantee the common path doesn't deliver.

Suggested easy follow-up (either option, both adaptor-only):
1. Unconditionally enqueue `cleanup_render_elements` in `on_cleanup` — the base class already no-ops when nothing was configured, so this is safe and makes the restore real; or
2. Narrow the docstrings on `_restore_arnold_aov_drivers` / `_configure_renderer_output` to state the restore only fires when render-element modification is enabled.

Fine to land this PR without it and track as a follow-up issue.

leongdl (dismissed):

src/deadline/max_adaptor/MaxClient/render_handlers/arnold_handler.py
ZainAallii: fixed

leongdl (approved):

leongdl:
# Review Guide

*A narrative walkthrough of this PR for reviewers: the problem, the code flow, and what to scrutinize.*

## 1. The Problem

Arnold (MAXtoA) AOVs did not survive animation rendering through the Deadline Cloud adaptor, for two independent reasons:

1. **Per-frame filename collisions.** Arnold writes one file per AOV driver, and the file's basename comes from that driver's `filenameSuffix` (e.g. `AOVs.exr`). There is no per-frame token, so every frame overwrote the previous frame's AOV files — only the last frame's AOVs survived. A hard blocker for animation.
2. **Scene-baked output path.** `AOV_Manager.outputPath` is authored on the artist's machine and points at their local filesystem, which is not writable (and not job-attachment-captured) on a worker.

The structural wrinkle that makes this PR non-trivial: Arnold AOVs live in their own subsystem (`renderer.AOV_Manager`, an `ArnoldAOVsManager` reference target), **not** in 3ds Max's standard `RenderElementManager`. So none of the existing render-element plumbing (the path V-Ray uses) applies — the handler has to drive the AOV manager directly through pymxs.

## 2. The Big Picture

All production logic is in one file: `arnold_handler.py` (+258). The mechanism, per session:

```
scene loads on worker
     │
per frame: adaptor calls handler._configure_renderer_output(output_name, output_dir, fmt)
     │   1. first frame only: cache original AOV_Manager.outputPath
     │      and every driver's original filenameSuffix (in driver order)
     │   2. force AOV_Manager.outputPath = job output dir
     │   3. rewrite each driver's filenameSuffix to
     │      "<output_name>_<original_suffix>"   (e.g. cam_render0001_AOVs)
     │      — empty/"undefined" suffix → "<output_name>_AOV<i>" safety tag
     │      — uniqueness enforced unconditionally via a per-frame set;
     │        collisions disambiguated with the driver index
     │   4. return False → framework still writes the beauty pass
     │        via rt.render(outputFile=...)
     ▼
session end: cleanup_render_elements()
     ├─ _restore_arnold_aov_drivers(): put back original outputPath +
     │    suffixes; try/finally guarantees the caches are cleared
     └─ super().cleanup_render_elements() — standard render-element restore,
          guaranteed to run even if the AOV restore blows up
```

Because `output_name` already contains the resolved frame number, injecting it into every driver's suffix is what makes each frame's AOV files unique.

## 3. What to read carefully

### The uniqueness algorithm (`_configure_renderer_output`)

This went through three iterations during review, each closing a collision class:
1. Original: empty suffix → collides with the beauty pass. Fixed with an `_AOV` tag.
2. `_AOV` tag → two empty-suffix drivers collide with *each other*. Fixed with `_AOV<index>`.
3. `_AOV<index>` → two drivers with the same *authored* suffix (both `AOVs`) still collide, and an authored suffix can equal another driver's safety tag (`AOV0`). Fixed with the final design: track composed names in a per-frame `used_suffixes` set and append the driver index on any collision, with a `while` loop guarding the pathological repeat-collision case.

Each iteration has a pinning test (`test_multiple_empty_suffix_drivers_do_not_collide`, `test_duplicate_authored_suffixes_do_not_collide`, `test_authored_suffix_colliding_with_safety_tag_is_disambiguated`).

### The cache/restore lifecycle

Originals are captured **once, on the first frame** — capturing per frame would cache the already-mutated value and compound suffixes across frames (`test_caches_original_suffix_only_once_across_frames` pins this, including that a later frame with a different dir doesn't re-capture). Restore happens in `_restore_arnold_aov_drivers`, and after review now covers *both* mutated pieces of state (suffixes **and** `outputPath` — the latter was initially missed). `try/finally` clears the caches even if a restore raises, and the one previously-unguarded pymxs access (`aov_mgr.drivers` during restore) is now wrapped so a failure can't escape and skip the standard render-element restore in `super()`.

Two documented assumptions to be aware of:
- **Driver list is assumed static per session** (cache is keyed by index/order). Fine for a render job; would break if drivers were added/removed mid-session — the comment says to switch to keying by object identity if that ever happens.
- **The cleanup hook may not run at all.** The adaptor only enqueues `cleanup_render_elements` when `enabled_modify_render_elements` is set, which a typical Arnold AOV job may not enable. The docstring explicitly accepts this: the restore is defensive only (the worker discards its scene copy), and the code guarantees it can't block the standard restore when the hook *is* enqueued.

### `check_renderer` robustness

`startswith("Arnold")` instead of `==`, so a future versioned MAXtoA class name (`Arnold_7_x_x`, V-Ray style) still routes to this handler instead of silently falling through to Default Scanline and losing the AOV wiring. Pinned by the new parametrized test case.

### `_normalize_mxs_str`

pymxs stringifies unset values as `"undefined"`, which is truthy. Without normalization the literal string `undefined` would leak into output filenames and the restore cache. `test_undefined_suffix_is_treated_as_empty` pins it.

### Return value: always `False`

This confused a reviewer, so worth stating plainly: the Arnold handler only redirects AOV drivers — the framework must still call `rt.render(outputFile=...)` for the beauty pass, which only happens when this returns `False`. (`VrayHandler` returns `True` in raw-output modes because V-Ray writes the beauty pass into its own container file.) The docstring was reworded to make the contrast unambiguous.

## 4. Review history highlights

- Maintainer feedback simplified the design: the defensive `_iter_drivers` wrapper was removed in favor of direct iteration, and the suffix cache went from `Dict[int, str]` to an ordered `List[str]`.
- An earlier `_apply_path_mapping` override (remapping `outputPath` at scene load) was flagged as dead work — `_configure_renderer_output` unconditionally overwrites `outputPath` before any render — and was ultimately **removed** from the diff.
- Verified end-to-end by the author: OpenJD runner locally (3 frames × 3 drivers, no collisions) and a Deadline Cloud submission (2-driver scene, per-frame logs confirm remapping and no compounding, all frames uploaded).

## 5. What to scrutinize / residual items

1. **Stale PR description.** The description's "What was the solution?" still lists `_apply_path_mapping` as part of the change, but it was removed from the final diff (a reviewer flagged this along with the missing path-remap tests). The description should be updated before merge so the merged commit message doesn't document a method that doesn't exist.
2. **Everything is mocked.** The tests exercise the logic thoroughly, but all pymxs interaction is `MagicMock` — whether `AOV_Manager.outputPath`/`filenameSuffix` writes behave as expected in real MAXtoA rests on the author's manual verification, not CI.
3. **Restore may silently never run** (the `enabled_modify_render_elements` gating above). Documented and accepted, but it means the "scene not polluted across sessions" guarantee is conditional.
4. **Suffix rewrite happens even for single-frame stills**, where the per-frame token is unnecessary — harmless, but output names change for existing still-image AOV workflows (`AOVs.exr` → `<output_name>_AOVs.exr`). Arguably an improvement (ties AOVs to the render), just worth knowing it's a behavior change for stills too.

## 6. Verdict framing

Tightly-scoped change: one handler file plus tests, no schema change, beauty pass untouched, non-AOV jobs unaffected. The failure mode it fixes (frames overwriting each other's AOVs) is strictly worse than any residual risk. The review loop demonstrably converged — each collision class the reviewers found was closed with a fix *and* a pinning test, and the maintainer simplification feedback was taken. The only pre-merge ask is item 1: refresh the stale description.

---
*Generated review guide — posted by @leongdl*

jericht (approved):

PR #272, opened 2026-07-22
feat: add 3dsmaxcmd command-line workflow

ZainAallii:
Adds a '3dsmaxcmd Render' submitter tab that builds a job bundle rendering the scene with 3dsmaxcmd.exe (the render server) instead of the standard 3dsmaxbatch adaptor. This lets network-licensed plugins such as Pencil+ (NTR) render without a watermark.

The task driver applies Deadline Cloud path mapping to the loaded scene via a pre-render MAXScript, then invokes 3dsmaxcmd per frame. The job gets the base dialog's Shared job settings (name/description) and Host Requirements, with an optional per-task run timeout. Generic template helpers are extracted into utilities/job_template_utils.

Network-licensed plugins such as Pencil+ (NTR) render with a watermark when rendered through the standard adaptor (`3dsmaxbatch.exe`), because it does not register as a render server. 

Added an opt-in "3dsmaxcmd Render" tab to the submitter that builds a job bundle rendering the scene with `3dsmaxcmd.exe` (the 3ds Max render server), which is what lets network-licensed (NTR) plugins render watermark-free. The change includes a new submitter tab, settings dataclass, bundle-creation callback, task driver, and job template. The task driver reads the session's Deadline Cloud path-mapping rules and generates a pre-render MAXScript that remaps the scene's asset and output paths to their session locations, then invokes `3dsmaxcmd` per frame. The job carries the base dialog's Shared job settings (name/description) and Host Requirements, plus an optional per-task run timeout. The generic template helpers (`load_job_template` / `inject_embedded_script`) were extracted into `utilities/job_template_utils` and re-exported from the V-Ray module so both workflows share one implementation.

Adds a new, opt-in render path. Default submitter behavior is unchanged unless the artist enables the "3dsmaxcmd Render" tab.

- Added 13 unit tests for the task driver's pure logic (path-mapping parse, pre-render MAXScript generation, command assembly, output sanitization). Full submitter unit suite passes (197 tests). `ruff` and `black` clean.
- Validated end-to-end on a Deadline Cloud service-managed fleet: submitted from the tab, path mapping applied, rendered on the worker via `3dsmaxcmd`, and output uploaded successfully. Verified on 3ds Max 2025 and 2027.

Inline: tab labels/tooltips, job-template parameter descriptions, and module/function docstrings (including the callback's asset-coverage contract).

- [ ] Yes
- [x] No

- [ ] Yes
- [x] No

No.

----


src/deadline/max_submitter/scripts/maxcmd_render.py
```diff
@@ -0,0 +1,295 @@
...
+    )
+
+    fn _remapPath p = (
+        local np, src, result
+        result = p
+        if p != undefined and p != "" then (
+            np = _normPath p
+            for pair in remaps while result == p do (
+                src = pair[1]
+                if (matchPattern np pattern:(src + "*")) then (
```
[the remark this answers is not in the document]
ZainAallii: Fixed this one.

src/deadline/max_submitter/scripts/maxcmd_render.py
```diff
@@ -0,0 +1,295 @@
...
+        os.path.dirname(os.path.abspath(__file__)), "maxcmd_prerender.ms"
+    )
+    with open(prerender_script, "w") as fh:
+        fh.write(build_prerender_maxscript(rules))
+
+    # Resolve the output target (directory + filename) if a name was provided.
+    output_path = ""
+    if output_file_name:
+        os.makedirs(output_dir, exist_ok=True)
+        output_path = normalize_path(os.path.join(output_dir, output_file_name))
```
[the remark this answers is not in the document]
ZainAallii: Verified on a multi-frame Deadline Cloud render — 3dsmaxcmd does append the frame number for single-frame (start == end) invocations, so there's no collision.

Because the driver always renders with an explicit range (-start:N -end:N), 3ds Max runs in sequence mode and inserts the padded frame number into the -outputName stem itself. The number comes from the frame being rendered, so parallel per-frame tasks each write a distinct file even though they all pass the same -outputName.

Confirmed with a 3-frame job (each frame its own task, all passing -outputName:cmdtest.jpeg): the outputs came back as cmdtest0001.jpeg, cmdtest0002.jpeg, cmdtest0003.jpeg (download reported cmdtest%04d.jpeg, 3 files, sequence 1-3) — plus the render-element pass line_###0001/0002/0003.jpg, all distinct, nothing overwritten.

So no per-frame token is injected in the driver — doing so would double-number since Max already pads (cmdtest.0001.0001.jpeg). Passing the plain stem is intentional and documented in build_maxcmd_command.

src/deadline/max_submitter/scripts/maxcmd_render.py
```diff
@@ -0,0 +1,382 @@
...
+        fh.write(prerender_body)
+
+    # Resolve the output target (directory + filename) if a name was provided.
+    # The filename is passed as a plain stem; 3ds Max inserts the per-frame
+    # number itself (sequence mode from -start/-end), so distinct frames don't
+    # collide. See build_maxcmd_command for details.
+    output_path = ""
+    if output_file_name:
+        os.makedirs(output_dir, exist_ok=True)
+        output_path = normalize_path(os.path.join(output_dir, output_file_name))
```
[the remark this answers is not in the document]
ZainAallii: Resolved in eb40be9. When output_file_name is blank (no -outputName), the pre-render script no longer relies on rule-based remap for the scene's baked rendOutputFilename. It now redirects the main render output into the job output directory — the same mechanism used for render elements:

if rendOutputFilename != undefined and rendOutputFilename != "" then (
    if jobOutputDir != "" then (
        rendOutputFilename = jobOutputDir + "/" + (filenameFromPath rendOutputFilename)
    ) else (
        rendOutputFilename = _remapPath rendOutputFilename
    )
    makeDir (getFilenamePath rendOutputFilename) all:true
)
The main pass now lands in the writable, path-mapped, output_directories-registered job output dir — no 0x3 abort, and the image is captured by job attachments. _remapPath stays only as a fallback for the non-cloud case with no job output dir; since maxcmd_submitter requires the output directory, jobOutputDir is always set for a real submission, so the redirect always applies. Blank filename ("use scene Render Setup output") stays supported.

Covered by test_main_output_redirected_to_output_dir and test_main_output_falls_back_to_rule_remap_without_output_dir, and verified on a multi-frame Deadline Cloud render (main output + Pencil+ Line pass both landed in the job output dir and were captured).

leongdl (dismissed):

leongdl:
# Review Guide

*A narrative walkthrough of this PR for reviewers: the problem, the code flow, and what to scrutinize.*

## 1. The Problem

Network-licensed (NTR) 3ds Max plugins — the motivating case is Pencil+ — check whether the host process is registered as a **render server**. The existing Deadline Cloud integration renders through the adaptor using `3dsmaxbatch.exe`, which does *not* register as one, so those plugins render with a watermark. `3dsmaxcmd.exe` (the command-line render server) does register, but using it means bypassing the adaptor entirely — and the adaptor is what normally handles path mapping, output routing, and renderer setup on the worker.

So this PR solves two problems:
1. Give artists an opt-in path to render via `3dsmaxcmd.exe` (watermark-free NTR plugins).
2. Re-implement, minimally, the adaptor services that `3dsmaxcmd` doesn't get for free: rewriting scene-baked absolute asset paths to session locations, and forcing outputs into a writable, job-attachment-captured directory.

Everything non-obvious in this diff is a consequence of #2.

## 2. The Big Picture: two halves, two machines

```
ARTIST MACHINE (submit time)
  submit_dialog.py ─ adds "3dsmaxcmd Render" tab, routes submission
       │
  maxcmd_tab.py ──── Qt tab UI (enable checkbox, frames, output, camera, timeout)
       │
  maxcmd_settings.py ─ dataclass + sticky-settings persistence per scene
       │
  maxcmd_submitter.py ─ builds the job bundle:
       │                 loads maxcmd_render_job_template.yaml,
       │                 injects scripts/maxcmd_render.py as an embedded file,
       │                 writes template/parameter_values/asset_references YAML
       ▼
  ──────────── job bundle ────────────► Deadline Cloud

WORKER (render time, per frame task)
  maxcmd_render.py (task driver, pure Python, no pymxs)
       1. parse argv (Param/Task/Session values arrive as arguments)
       2. read Session.PathMappingRulesFile JSON
       3. generate maxcmd_prerender.ms (MAXScript) with the rules baked in
       4. exec: 3dsmaxcmd -start:N -end:N -outputName:... -preRenderScript:pre.ms scene.max
       5. inside Max, the pre-render script remaps bitmap/XRef paths and
          redirects outputs to the job output dir, then the frame renders
       6. driver streams sanitized output; on failure dumps the generated
          script and Max's network log for diagnosis
```

Suggested reading order: `maxcmd_render_job_template.yaml` → `scripts/maxcmd_render.py` → `maxcmd_submitter.py` → `ui/submit_dialog.py` → the rest.

## 3. File-by-file walkthrough

### `job_templates/maxcmd_render_job_template.yaml` (new) — the contract

OpenJD template: one `Render` step whose task parameter space is the frame range, so **each frame is its own task** (`Frame` INT with `range: '{{Param.Frames}}'`). The task driver is an embedded file with a `INJECT_MAXCMD_RENDER_SCRIPT` placeholder that the submitter substitutes at bundle time.

Worth pausing on: parameters are passed as **argv** (`--scene-file '{{Param.SceneFile}}'` etc.), not interpolated into the Python source. An earlier revision interpolated into `r"..."` raw-string literals; a Windows path ending in `\` or a value containing `"` would have been a `SyntaxError` on the worker. Fixed at the root.

### `scripts/maxcmd_render.py` (new, 411 lines) — the heart of the change

Runs per task on the worker. Deliberately dependency-free (stdlib only) since it executes under plain `python`, not inside Max. Helpers are pure and unit-tested; `main()` is the only impure orchestration.

- `load_path_mapping_rules()` — reads Deadline's session path-mapping JSON into `(source, dest)` pairs. Missing/corrupt file degrades to "no remapping" (correct for storage-profile workers that see original paths).
- `build_prerender_maxscript()` — **the trickiest code in the PR**: Python generating MAXScript. Three things to check carefully:
  - `_remapPath`: case/slash-insensitive prefix match with a **path-boundary guard** (`tail == "" or tail[1] == "/"`) so a rule for `L:/tex` can't capture `L:/textures/...`.
  - Every section is wrapped in MAXScript `try/catch` — an unhandled pre-render error makes 3dsmaxcmd abort the whole render with an opaque "unexpected exception" message, so failures are logged-and-continued per asset class instead.
  - **Output redirect logic**: rule-based remapping can't work for outputs, because Deadline only creates path-mapping rules for *input* roots. So the main output (`rendOutputFilename`) and every render element are redirected into `jobOutputDir` keeping just their basename, with `makeDir` (avoids the 0x3 "path not found" abort). `_remapPath` remains only as the no-output-dir fallback.
- `build_maxcmd_command()` — assembles the CLI. Output name is passed as a **plain stem**: because `-start:N -end:N` puts Max in sequence mode, Max itself pads the frame number (`render0001.exr`), so per-frame tasks don't collide. Empirically verified on a 3-frame cloud render; `test_output_name_passed_verbatim_without_frame_token` pins this.
- `run_maxcmd()` — streams subprocess output, laundering it to ASCII. 3dsmaxcmd emits non-UTF-8 banner bytes that would crash the worker's strict UTF-8 stdout reader and falsely fail a successful render.
- `dump_max_network_log()` — on failure, prints the generated MAXScript and the per-user `Max.log`, since 3dsmaxcmd's stdout is uselessly terse on error.

### `maxcmd_submitter.py` (new) — bundle construction

`on_create_maxcmd_job_bundle_callback()`: validates (scene saved, frames set, **output directory required** — scene-baked output paths don't exist on workers and wouldn't be captured), loads the template, injects the driver script, writes the three bundle YAMLs, applies Host Requirements per step, and injects the optional per-task `timeout` on `onRun`.

Read the docstring — it's honest about scope: asset coverage is whatever the base dialog's auto-detection plus Job Attachments tab provides, and **path-mapping only covers bitmaps and XRefs**. VRayProxy/Alembic/IES/OSL paths are uploaded but not rewritten, so scenes using them aren't fully supported yet.

### `ui/submit_dialog.py` — routing

Adds the tab unconditionally (renderer-agnostic, unlike the V-Ray tab) and checks it **first** in `_on_create_job_bundle_wrapper` — maxcmd takes precedence over V-Ray export if both are somehow enabled. Defaults output path/filename from the scene's `rendOutputFilename`, then lets sticky settings override.

### `ui/maxcmd_tab.py` (new) — plain Qt

Enable checkbox gating a settings group; mirrors the standard tab's timeout row. Job name/description/priority intentionally omitted (owned by Shared job settings). Low review risk.

### `maxcmd_settings.py` (new) — sticky settings

Dataclass with `metadata={"sticky": True}` fields persisted to `<scene>.deadline_maxcmd_settings.json`, same pattern as the V-Ray settings. Deliberate choice: **output path is not sticky** — it's re-derived from the scene each time so a stale value can't mask a changed Render Setup output.

### `utilities/job_template_utils.py` (new) + `vrscene_job_submission.py` — refactor

`_load_job_template`/`_inject_embedded_script` extracted verbatim from the V-Ray module into a renderer-agnostic utility and re-exported under their old private names, so V-Ray call sites are untouched. Pure move; existing V-Ray tests pass unmodified.

### Tests (+312)

`test_maxcmd_render.py` loads the driver by file path (it's an embedded script, not an importable package) and covers the pure helpers: rules parsing, MAXScript generation (including the path-boundary guard and both output-redirect branches), non-UTF-8 output resilience, command assembly (pins no frame token), and the argv-safety cases (trailing backslash, embedded quote). Note what's *not* tested: the generated MAXScript is asserted by string content, not executed — its actual behavior in Max is validated by the author's manual end-to-end runs on Max 2025/2027.

## 4. What to scrutinize / known limitations

1. **Render-element basename collision (open):** elements are flattened into `jobOutputDir` by basename. Two elements authored as `.../beauty/render.exr` and `.../line/render.exr` would silently overwrite each other. Arguably within the documented "full render-element support deferred to follow-up" scope, but it's a silent-data-loss mode — worth an explicit ack or a name-disambiguation tweak in the follow-up.
2. **MAXScript-in-Python-f-string escaping** (`_escape_maxscript_string` and the template literal) is the most fragile surface. Tests check the emitted text but nothing executes it; regressions here only show up on a real worker.
3. **Asset-type coverage gap** is documented, not enforced: a scene with VRayProxies submits fine and fails (or renders wrong) on the worker. No submit-time warning.
4. **Worker behavior claims** (Max sequence-mode padding, 0x3 aborts, non-UTF-8 banners) rest on manual verification — documented in comments/tests, but not reproducible in CI.

## 5. Verdict framing

Opt-in feature: default submitter behavior is unchanged unless the artist enables the tab, so blast radius is limited to the new workflow. The design correctly keeps the worker driver pure/testable and pushes all Max-API interaction into generated MAXScript. The review history shows every substantive automated-review finding (path-boundary bug, raw-string injection, blank-output-filename capture, frame numbering) was either fixed or empirically refuted. The one live question is the render-element collision in item 1 above.

---
*Generated review guide — posted by @leongdl*

leongdl (commented):
Following up on the review guide: posting the small (<50-line) fixes that would close the remaining open findings. The main one is the render-element basename collision flagged by the automated review — a contained fix in the generated MAXScript, sketched inline below. The other two are one-line deletions to clear the CodeQL findings.

src/deadline/max_submitter/scripts/maxcmd_render.py
```diff
@@ -0,0 +1,411 @@
...
+        reMgr = maxOps.GetCurRenderElementMgr()
+        if reMgr != undefined then (
+            for reIdx = 0 to (reMgr.NumRenderElements() - 1) do (
+                reFn = reMgr.GetRenderElementFilename reIdx
+                if reFn != undefined and reFn != "" then (
+                    -- Redirect to the job output directory (keeping the
+                    -- element's own filename, so frame-number tokens survive)
+                    -- when one is set; otherwise fall back to rule-based remap.
+                    if jobOutputDir != "" then (
+                        reMapped = jobOutputDir + "/" + (filenameFromPath reFn)
```
leongdl: Proposed fix for the basename-collision finding above (elements like `.../beauty/render.exr` and `.../line/render.exr` flattening to the same `jobOutputDir/render.exr`). Same shape as the PR #266 disambiguation: track composed names, only rename on collision — so names are unchanged in the common case and your verified end-to-end runs stay representative.

In the generated MAXScript, seed a used-names list with the main output's basename (so an element can't clobber the redirected beauty pass either), declared with the other locals at the top of the block:

```maxscript
local usedNames = #()
if rendOutputFilename != undefined and rendOutputFilename != "" then (
    append usedNames (toLower (filenameFromPath rendOutputFilename))
)
```

Then replace this flat-basename redirect with:

```maxscript
if jobOutputDir != "" then (
    local baseName = filenameFromPath reFn
    if (findItem usedNames (toLower baseName)) != 0 then (
        -- collision: insert the element index before the extension so
        -- sequence-mode frame padding still lands in the right place
        baseName = (getFilenameFile baseName) + "_re" + (reIdx as string) + (getFilenameType baseName)
    )
    append usedNames (toLower baseName)
    reMapped = jobOutputDir + "/" + baseName
) else (
    reMapped = _remapPath reFn
)
```

Notes:
- `toLower` comparison because Windows filesystems are case-insensitive — `Render.exr` vs `render.exr` collide on disk.
- Index goes **before** the extension (`render_re1.exr`), so 3dsmaxcmd's frame padding composes correctly (`render_re10001.exr`).
- `reIdx` is unique per element, so a single pass suffices — no while-loop needed since the tag embeds the literal basename.
- ~15 lines in the script plus 2–3 string-assertion unit tests in the existing style. Worth one re-verification on a real worker since nothing executes the generated MAXScript in CI.

src/deadline/max_submitter/maxcmd_settings.py
```diff
@@ -0,0 +1,87 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+"""3ds Max Command-Line (3dsmaxcmd) Render Settings Data Class."""
+
+import dataclasses
+import json
+from dataclasses import dataclass, field
+from pathlib import Path
+
+import pymxs  # noqa
```
leongdl: One-line cleanup to clear the CodeQL `Unused import` finding: this bare `import pymxs` is redundant — the `from pymxs import runtime as rt` two lines down already imports the module (and provides the same fail-fast-outside-Max behavior).

```suggestion
```

src/deadline/max_submitter/maxcmd_submitter.py
```diff
@@ -0,0 +1,147 @@
...
+absolute paths baked into the scene resolve to their session locations on the
+worker.
+"""
+
+import logging
+import os
+from pathlib import Path
+from typing import Any, Optional
+
+import pymxs  # noqa
```
leongdl: Same CodeQL `Unused import` finding here. Nothing in this module references `pymxs` or `rt` directly, and the module already fails fast outside Max via its imports of `maxcmd_settings` / `ui.submit_dialog` (both import pymxs). If the import is intentionally kept as an explicit in-Max guard, a short comment saying so would let the CodeQL suppression be justified; otherwise it can just be dropped:

```suggestion
```

src/deadline/max_submitter/scripts/maxcmd_render.py
leongdl: **Gap: V-Ray VFB raw output paths are not redirected.**

The pre-render script covers bitmap textures, XRefs, `rendOutputFilename`, and `RenderElementManager` elements — but V-Ray's VFB raw output lives in none of those. It's renderer-owned state (`output_saveRawFile` / `output_rawFileName`, and the split-channel variants `output_splitgbuffer` / `output_splitfilename`), baked to the artist's local disk, and V-Ray writes it itself, bypassing Max's output plumbing entirely. This is exactly the state the adaptor's `VrayHandler` exists to manage, and none of that reaches the maxcmd path.

So a V-Ray scene with VFB raw output submitted through this tab either fails the write mid-render (path doesn't exist on the worker) or writes outside `output_directories` and the frames are silently never captured.

The fix is another try/catch section mirroring the `rendOutputFilename` redirect above — something like:

```maxscript
-- V-Ray VFB raw output: renderer-owned path, invisible to rendOutputFilename
-- and the render element manager. Redirect to the job output dir.
try (
    local r = renderers.current
    if (isProperty r #output_rawFileName) and r.output_saveRawFile then (
        if r.output_rawFileName != undefined and r.output_rawFileName != "" then (
            r.output_rawFileName = jobOutputDir + "/" + (filenameFromPath r.output_rawFileName)
            makeDir (getFilenamePath r.output_rawFileName) all:true
        )
    )
    if (isProperty r #output_splitfilename) and r.output_splitgbuffer then (
        if r.output_splitfilename != undefined and r.output_splitfilename != "" then (
            r.output_splitfilename = jobOutputDir + "/" + (filenameFromPath r.output_splitfilename)
            makeDir (getFilenamePath r.output_splitfilename) all:true
        )
    )
) catch (
    format "maxcmd output map: vray raw output remap error (continuing): %\n" (getCurrentException())
)
```

Notes:
- `isProperty` guards make it a no-op for every non-V-Ray renderer, so no renderer detection needed.
- The redirected basenames should feed the same used-names collision list proposed for the render-element fix, so a raw output named like an element doesn't clobber it.
- Please verify the property names against `VrayHandler` in this repo (it's the authoritative list of what the adaptor already remaps) and the V-Ray version on the fleet.
- Only `jobOutputDir != ""` branch shown; the rule-based `_remapPath` fallback applies as elsewhere.

If this lands in the follow-up PR instead, the tab tooltip / limitation note should say explicitly: *"V-Ray VFB raw output is not redirected — use the standard Output Filename field or the adaptor workflow for V-Ray."* Right now that failure is silent, which is the worst version of it.

leongdl (commented):
One more gap worth capturing while we're closing the small ones: V-Ray VFB raw output is not redirected by the pre-render script. Inline comment below has the details and a ~25-line fix in the same pattern as the existing output blocks.

leongdl:
Nevermind on the VRay comments - those can be deferred as a second CR. Right now we want 3dsmaxcmd support first.

leongdl (dismissed):

src/deadline/max_submitter/maxcmd_settings.py
ZainAallii: done

jericht:
Read through David's review guide and followed the recommended order (job template → `scripts/maxcmd_render.py` → `maxcmd_submitter.py` → `ui/submit_dialog.py` → the rest). Traced the four focus areas he called out:

1. **Render-element basename collisions** — the resolution converges. `usedNames` is seeded from the beauty pass, and each disambiguated name is appended before the next element iterates, so a disambiguated name can't collide with a later element's original. The remaining gap is the one already noted: the explicit `-outputName` basename isn't seeded into `usedNames`, so an element whose filename stem exactly matches the artist's chosen output name would still clash. Agree that's follow-up material rather than a blocker.

2. **MAXScript-in-Python escaping** — holds up. `normalize_path` converts backslashes to forward slashes before `_escape_maxscript_string` runs, so `"` is the only special character left and it's escaped. I couldn't construct a path or camera name that breaks out of the generated string literals.

3. **Asset-type coverage** — confirmed the callback doesn't scan the scene itself and relies on the artist's attachments. The docstring is explicit about it, which I think is the right call for an opt-in workflow, but it does mean a missing IES file or proxy surfaces as a render-time failure rather than a submit-time warning.

4. **Worker behavior / failure detection** — verified by reading rather than running: 3dsmaxcmd's exit code propagates `run_maxcmd` → `main()` → `sys.exit(main())`, stderr is merged into stdout, and the command is passed as an argv list with no `shell=True`. A failed render fails the task rather than silently succeeding.

No blockers from me.

jericht (dismissed):

AlexTranAmz (dismissed):

src/deadline/max_submitter/scripts/maxcmd_render.py
```diff
@@ -0,0 +1,437 @@
...
+        if rendOutputFilename != undefined and rendOutputFilename != "" then (
+            if jobOutputDir != "" then (
+                rendOutputFilename = jobOutputDir + "/" + (filenameFromPath rendOutputFilename)
+            ) else (
+                rendOutputFilename = _remapPath rendOutputFilename
+            )
+            makeDir (getFilenamePath rendOutputFilename) all:true
+            -- Claim the (redirected) beauty-pass basename so a render element
+            -- can't clobber it.
+            append usedNames (toLower (filenameFromPath rendOutputFilename))
```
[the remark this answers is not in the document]
ZainAallii: Fixed. The resolved `-outputName` filename is now passed into `build_prerender_maxscript` and seeded into `usedNames` at the top of the generated MAXScript (as `cmdOutputName`). This covers the common case where `rendOutputFilename` is empty (artist using the explicit Output Filename field rather than the scene's Render Setup output). Both seeds coexist: `cmdOutputName` protects the `-outputName` beauty pass, and the existing `rendOutputFilename` seed protects the scene-baked fallback. A render element whose basename collides with either now triggers the `_re<index>` disambiguation instead of silently overwriting.

seant-aws (approved):

jericht (approved):

wyongzhi (approved):
Non-blocking: my understanding is each task will launch a fresh `3dsmaxcmd.exe` and reloads the scene (one frame per task), so startup + scene-load cost is paid per frame rather than per session like the adaptor path. Have we considered this impact to the total time of rendering?

AlexTranAmz (approved):

PR #293, opened 2026-09-14
fix(vray): resolve tile files when V-Ray appends its own frame number

ZainAallii:
## What was wrong

`tile_render.py` bakes a padded frame number into the `-imgFile` name and also
passes `-frames`, which puts V-Ray in sequence mode so it inserts its own padded
frame number as well. The file written to disk is therefore:

```
_tile{N}_{output name}.{frame}.{frame}{ext}
```

while `tile_merge.py` looked only for the single-numbered form. Every
`MergeRegions` task failed with `ERROR: Missing tile` and no merged frame was
ever produced — even though the tiles rendered correctly, uploaded correctly,
and were synced into the merge task's session by job attachments.

This is the same double-numbering trap already documented in
`maxcmd_render.py`'s `build_maxcmd_command`, where injecting a frame token
alongside an explicit range produces `render0001.0001.exr`.

## What this does

Resolves tiles through a new `find_tile()` that tries, in order:

1. the name as requested,
2. the doubled form V-Ray actually writes,
3. any file for that tile whose dot-delimited numeric tokens include the frame.

Tokens are compared numerically as whole tokens, so frame 1 (`0001`) cannot
match frame 10001 (`10001`), and a padding width other than four still resolves.

`tile_render.py` is deliberately left untouched. Fixing the merge side keeps
working any V-Ray build or configuration that honours the requested filename
verbatim, rather than betting on one numbering behaviour.

On a miss, the error now lists every `_tile*` file present in the output
directory, so a naming mismatch is diagnosable from the task log instead of
requiring a separate investigation.

## Testing

- 14 unit tests covering both filename spellings, an output name containing
  spaces, an unpadded frame token, EXR, a full 2x2 grid resolving to four
  distinct files, and the negative cases: wrong frame, wrong tile index, and the
  frame 1 vs frame 10001 substring trap.
- End-to-end through the real OpenJD runtime using the same embedded-script
  mechanism the job template uses, with a **negative control**: the previous
  code reproduces the exact `Missing tile` failure, and this change produces a
  correct merge from the same inputs.
- Merged output verified quadrant by quadrant against its source tiles, not just
  checked for existence.
- On a farm render, two tile jobs went from total merge failure to 48/48 tasks
  succeeded.

## Note on tile appearance

An individual `_tile*` file is largely black by design — V-Ray `-region` writes
a full-size frame with only that region's pixels populated (measured: 0.00%
non-black in the other three quadrants of a 2x2). The merged frame is the
artifact to inspect.


src/deadline/max_submitter/scripts/tile_merge.py
```diff
@@ -39,6 +39,54 @@ def normalize_path(p):
...
+    verbatim instead.
+
+    Rather than depend on one of those behaviours, resolve the tile by looking
+    for the known spellings first and then falling back to any file for this
+    tile whose dot-separated numeric tokens include this frame. That last step
+    also covers a padding width other than 4.
+
+    Returns the resolved path, or None when nothing matches.
+    """
+    import glob
```
leongdl: Please move imports to the top.

src/deadline/max_submitter/scripts/tile_merge.py
```diff
@@ -157,20 +205,34 @@ def main():
...
+            else:
+                tile_paths.append(tile_file)
 
     if missing > 0:
         print(f"ERROR: {missing} tile(s) missing")
+        # List what is actually present so a naming mismatch is diagnosable from
+        # the task log instead of requiring a separate investigation.
+        try:
+            present = sorted(name for name in os.listdir(output_dir) if name.startswith("_tile"))
+            print(f"Found {len(present)} tile file(s) in {output_dir}:")
```
leongdl: nit for the future, maybe use a logger.

test/unit/deadline_submitter_for_3ds_max/test_tile_merge.py
```diff
@@ -0,0 +1,112 @@
...
+
+def _load_tile_merge():
+    spec = importlib.util.spec_from_file_location("tile_merge_under_test", str(_SCRIPT))
+    assert spec is not None and spec.loader is not None
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
+tile_merge = _load_tile_merge()
```
leongdl, line 43: **Blocking: importing `tile_merge.py` at module scope leaks `sys.path` into the rest of the pytest session.** This is a test-flakiness vector, not a style point.

This line runs at collection time, and `tile_merge.py` has import-time side effects — it creates `<tmpdir>/deadline_pip_packages` and prepends it to `sys.path`, in the block above `ensure_package`. Measured by loading the file exactly the way this test does:

```
dir exists before import : False
sys.path[0] before       : /private/tmp
dir exists after import  : True
sys.path[0] after        : /var/folders/.../T/deadline_pip_packages
mutated sys.path         : True
```

The failure this produces will not reproduce on a clean machine. `ensure_package` pip-installs Pillow, numpy and OpenEXR into that same directory when the script runs for real. So any host that has ever executed a tile merge — a developer box, a reused CI runner — has those packages sitting in a directory that this test file now places *ahead of* the venv on `sys.path`, for every test collected afterwards, not just the ones in this file. Whichever test imports PIL or numpy next silently resolves it from tmp at whatever version was installed back then. Order-dependent, host-dependent, and nothing in the failure output points here.

Restoring `sys.path` around `exec_module` closes it and keeps the module-scope load:

```python
def _load_tile_merge():
    spec = importlib.util.spec_from_file_location("tile_merge_under_test", str(_SCRIPT))
    assert spec is not None and spec.loader is not None
    module = importlib.util.module_from_spec(spec)
    saved_path = list(sys.path)
    try:
        spec.loader.exec_module(module)
    finally:
        sys.path[:] = saved_path
    return module
```

(needs `import sys` alongside the existing imports.)

The `makedirs` still runs, but an empty directory in tmp is harmless once it is off `sys.path`.

leongdl (commented):
One blocking item on the test file, detailed inline. The `tile_merge.py` change itself looks correct — I confirmed the `-imgFile` / `-frames` double-numbering premise against `tile_render.py`, and the whole-token numeric comparison in tier 3 does keep frame 1 from matching frame 10001.

leongdl (changes requested):

test/unit/deadline_submitter_for_3ds_max/test_tile_merge.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+def _load_tile_merge():
+    spec = importlib.util.spec_from_file_location("tile_merge_under_test", str(_SCRIPT))
+    assert spec is not None and spec.loader is not None
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module
+
+
```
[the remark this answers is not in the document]
ZainAallii: fixed.

test/unit/deadline_submitter_for_3ds_max/test_tile_merge.py
ZainAallii: done, resolving conversation.

